PR #1, opened 2023-05-05
Upload objects and visualization

IvanPombo:
This PR serves to upload two possible objects to start playing (motorbike.obj and birdy.obj) with different quantity of points in the mesh.

Example of the visualization:
Screenshot 2023-05-05 at 17 00 38

Moreover, I also send you a visualization script, which not only provides the above visuals, but already gives some methodology on how to upload the data and the mesh.

In particular, `mesh.points` can be turned into a numpy array for ease of handling the points.

### DIFF
--- a/data/visualization.py
+++ b/data/visualization.py
@@ -1,0 +1,24 @@
+from absl import app, flags, logging
+import pyvista as pv
+
+FLAGS = flags.FLAGS
+
+flags.DEFINE_string("input_object", "data/blackbird.obj",
+                    "File path of the object to be visualized")
+
+def main(_):
+
+    logging.info("Loading the object from %s", FLAGS.input_object)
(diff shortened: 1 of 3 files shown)


sarmento:
Not sure if obj files should be in the repo? (or at least in the same dir as the code).
Are there IP rights on these? Or: are "open" libs of these that people can just download?

@hpenedones : wdyt?

data/visualization.py
```diff
@@ -0,0 +1,26 @@
+"""Upload an object and visualize it using pyvista."""
+from absl import app, flags, logging
+import pyvista as pv
```
fabiocruz, line 3: Add `pyvista` to the package requirements.

fabiocruz (approved):

PR #2, opened 2023-05-12
implementation of loading and processing pipeline

VictorGillioz:


datatypes/graph.py
```diff
@@ -0,0 +1,17 @@
+"""Defines custom data classes"""
```
IvanPombo, line 1: Try to be more specific in this early description.

datatypes/graph.py
```diff
@@ -0,0 +1,17 @@
...
+
+from dataclasses import dataclass
+import numpy as np
+
+
+@dataclass
+class EdgeSet:
```
IvanPombo, line 8: Add a one-liner describing this data class and its features. 

What is the sender and receiver? Does the graph contains edges with directions?

datatypes/graph.py
```diff
@@ -0,0 +1,17 @@
...
+    features: np.ndarray
+    senders: np.ndarray
+    receiver: np.ndarray
+
+
+@dataclass
+class Graph:
```
IvanPombo, line 15: One-liner :)

utils/data_processing.py
```diff
@@ -0,0 +1,44 @@
+"""Process and format the mesh data for our model"""
+
+from datatypes.graph import EdgeSet
+from datatypes.graph import Graph
+
+import numpy as np
+
+
+def _triangles_to_edges(faces):
```
IvanPombo, line 9: Are we completely sure that cells are defined by triangles? Can they be quadrilaterals?

utils/data_processing.py
```diff
@@ -0,0 +1,44 @@
...
+
+    edges = np.concatenate([faces[:, 0:2], faces[:, 1:3], faces[:, [0, 2]]],
+                           axis=0)
+
+    edges = np.sort(edges, axis=1)
+    unique_edges = np.unique(edges, axis=0)
+
+    senders, receivers = unique_edges[:, 0], unique_edges[:, 1]
```
IvanPombo, line 17: So this means that senders are the minor index and the receiver the higher index (by the sorting above). Is that it?

utils/data_processing.py
```diff
@@ -0,0 +1,44 @@
...
+
+    return (np.concatenate([senders,
```
IvanPombo, line 19: Ok, I see the point here. Does this structure arise from the literature you have seen?

I see the point, I just want to be sure if it is truly necessary...

IvanPombo (commented):
This is starting to take shape! I just left some minor comments and questions to grasp every concept and make sure that we are on the right path. 

A general suggestion is to provide a description in each function, explaining the args and returns. This will allow me to provide feedback faster and for others checking the repo to understand everything from scratch! Don't forget it is open-source :) 

If you want, you can also start describing the model we are implementing in the ReadMe. Later on, this can be useful for a paper or a presentation. 

This is are my first of comments! Let's start from here, and, I will provide a second round afterwards!!
Keep it up!

utils/data_processing.py
VictorGillioz: I worked with triangle cells because this is the objects I was provided with, but we can adapt the code later on to consider other shapes

utils/data_processing.py
VictorGillioz: This sorting allows to quickly identify duplicated edges

utils/data_processing.py
VictorGillioz: Yes, this follows MeshGraphNets implementation. Edges are present twice, once in each direction. I believe it is coherent with the message passing implementation in their model but I will make sure later if this is indeed necessary

utils/data_processing.py
IvanPombo: Have you checked the "bird.obj"? It seems to be quadrilateral.

To be more general, the appropriate way is to use the adjacency matrix and generate the edges from there. With Graph Networks we are not conflicted with type of mesh as in classical simulators, thus we can go for a more general methodology.

utils/data_processing.py
IvanPombo: Alright! It was just to be clear.

utils/data_processing.py
VictorGillioz: I could not load bird.obj with meshio because there is a problem with the its texture, but I think .obj files are always triangle meshes.

VictorGillioz: For the adjacency matrix, we rather encode the edges as a list here. The high number of nodes and low connectivity probably justify it. With quadrilateral meshes we should be able to easily add the 4th edge to this list.

IvanPombo: With pyvista you can read the mesh and see that it is quadrilateral. Obj supports polygonal meshes, but it might be the encoding that meshio is not reading correctly! We will make sure that with the dataset we can read everything independently of the mesh type.

With general meshes we can encode it has a list as well!

datatypes/graph.py
```diff
@@ -0,0 +1,21 @@
+"""Data classes used by our model"""
```
IvanPombo, line 1: Suggestion: Data classes for Graph Neural Network model.

IvanPombo: Let's describe here also the type of structure. That is, how the Graph class is defined with a list of features, senders and receivers for each point.

main.py
```diff
@@ -0,0 +1,36 @@
+"""The main file to run our code"""
```
IvanPombo, line 1: Suggestion: Script to load mesh data from an obj file to a graph.

IvanPombo (approved):
Here it goes another round of feedback! I think after the change with the meshes you can merge and start working on the pressure features.

PR #4, opened 2023-05-17
moved files for absolute import

VictorGillioz:
I moved the code files inside the "meshnets" folder and created a \_\_init\_\_.py file inside it. This offers a cleaner structure of the repo and allows absolute imports inside the python files.


IvanPombo:
Have you tested this on your side? I am not sure if it won't need `__init__.py` files inside the subfolders to work!

VictorGillioz:
It works fine without it, I think python takes care of it implicitly since 3.3. I still included it as it seems to be best practice in some cases but I can remove it if you think it is not needed.

IvanPombo:
> It works fine without it, I think python takes care of it implicitly since 3.3. I still included it as it seems to be best practice in some cases but I can remove it if you think it is not needed.

Let's keep it anyway!

IvanPombo (approved):
Lgtm!

PR #5, opened 2023-05-19
Geometric data

VictorGillioz:
I replaced the custom Graph type by torch_geometric.data.Data. In our case, it is not necessary to have custom datatypes and will simplify the pipeline down the road.


AugustoPeres (approved):
LGTM. Thanks @VictorGillioz

PR #6, opened 2023-05-19
torch_geometric.data.Batch as input to model

VictorGillioz:
Replaced custom Graph data by torch geometric Batch as input to the model


IvanPombo (approved):
LGTM!

PR #7, opened 2023-05-25
Identity encoder decoder implementation

VictorGillioz:
Skeleton for encoder and decoder torch modules. I would appreciate feedback on the following:

Do you think it should return the predicted tensor like right now or instead the graph batch with modified node features ?
Should  this code be added to the main model file or it will make the file too large once the processor is also implemented ?
Should I code the MLPs individually for the encoder/decoder or make a common method s.a. make_mlp(widht_list) ?

### DIFF
--- a/meshnets/modules/decoder.py
+++ b/meshnets/modules/decoder.py
@@ -1,10 +1,29 @@
 """"Define the GraphDecoder class."""
 
+import torch
+from torch.nn import Identity
+from torch_geometric.data import Batch
 
-class GraphDecoder:
 
-    def __init__(self, *args):
-        pass
+class GraphDecoder(torch.nn.Module):
+    """Decoder for graphs.
+    
+    Decode latent graphs to an output Tensor corresponding to predicted
+    node features by applying an MLP decoder to each node features.
+    """
 
-    def __call__(self, *args):
-        pass
+    def __init__(self, output_size, latent_size, num_mlp_layers):
+        """Initialize the node features decoder given the output size,
+        latent size and number of MLP layers."""
+        super().__init__()
+
+        # TODO(victor) : implement decoder
+        self.node_decoder = Identity(output_size, latent_size, num_mlp_layers)
+
+    def forward(self, graph: Batch) -> torch.Tensor:
+        """Decode the latent node features of a batch of graphs
+        to output node features.
+        
+        Return the output Tensor."""
+
+        return self.node_decoder(graph.node_features)
--- a/meshnets/modules/encoder.py
+++ b/meshnets/modules/encoder.py
@@ -1,10 +1,36 @@
 """"Define the GraphEncoder class."""
 
+import torch
+from torch.nn import Identity
+from torch_geometric.data import Batch
 
-class GraphEncoder:
 
-    def __init__(self, *args):
-        pass
+class GraphEncoder(torch.nn.Module):
+    """Encoder for graphs.
+    
+    Encode graphs to latent graphs by applying an MLP encoder to each
+    node features and a second MLP encoder to each mesh features.
+    """
 
-    def __call__(self, *args):
-        pass
+    def __init__(self, node_feats_size, mesh_feats_size, latent_size,
+                 num_mlp_layers):
+        """Initialize the node features encoder and the mesh features encoder
+        given the features sizes, latent size and number of MLP layers."""
+        super().__init__()
+
+        # TODO(victor) : implement encoders
+        self.node_encoder = Identity(node_feats_size, latent_size,
+                                     num_mlp_layers)
+        self.edge_encoder = Identity(mesh_feats_size, latent_size,
+                                     num_mlp_layers)
+
+    def forward(self, graph: Batch) -> Batch:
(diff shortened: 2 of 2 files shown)


sarmento (commented):
I have not a strong opinion now about these questions, I think it's all reasonable for now.

What I would like to see instead (maybe on the next PR) is to have all this Ray-enabled, in a way that you can start experimenting at scale even with very simple models.

Also, check with @AugustoPeres how much of this can be "moved" to the infra-structure package he is preparing so that you focus more and more on the definition of the data structures and  model, and less on the mechanics of training / scaling etc...

Thank you!

AugustoPeres:
Some high level comments:

> Do you think it should return the predicted tensor like right now or instead the graph batch with modified node features ?

If by predicted tensor you mean the output of the encoder or decoder I think that it is more standard to output tensors instead of the graph object. Why, because, by looking at the [source coder](https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/nn/conv/gcn_conv.html#GCNConv) of the base `MessagePassing` layer, we can see that it can receive pretty much whatever you want. And you could always have your own layer:

```python
class MyLayers(MessagePassing):
    def forward(self, graph):
        node_features = graph.x
        edge_attr = graph.edge_attr
```

However, there are already several layers implemented. Such as [GCNConv](urhttps://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.nn.conv.GCNConv.html#torch_geometric.nn.conv.GCNConv) that is of the form:

```python
def forward(self, x: Tensor, edge_index: Adj,
                edge_weight: OptTensor = None) -> Tensor:
```

Thus, It would be easier to chain in the model if previous layers where already outputing these types of tensors.


> Should this code be added to the main model file or it will make the file too large once the processor is also implemented ?

I usually prefer to have layers together in files.

> Should I code the MLPs individually for the encoder/decoder or make a common method s.a. make_mlp(widht_list) ?

Not sure of I understood but I don't think we need a method to create an MLP. We can just have an mlp that receives, in its `__init__` the `width_list` and then use:

```python
class Encoder(MessagePassing):
    def __init__(self, width_list, ...):
        self._mlp = MLP(width_list)
```


Additionally DLSP has just made public the [repository](https://github.com/DeepLearningPT/dlsp-workshop) for the code quest focusing on solving SPH with graph nets public. You can take a look at an Encoder -> Processor -> Decoder architecture there

meshnets/modules/decoder.py
```diff
@@ -1,10 +1,29 @@
...
-        pass
+    def __init__(self, output_size, latent_size, num_mlp_layers):
+        """Initialize the node features decoder given the output size,
+        latent size and number of MLP layers."""
+        super().__init__()
+
+        # TODO(victor) : implement decoder
+        self.node_decoder = Identity(output_size, latent_size, num_mlp_layers)
+
+    def forward(self, graph: Batch) -> torch.Tensor:
```
AugustoPeres, line 23: I am assuming that the decoder will appear after several layers of other messaging passing rounds. Therefore, since the output of [most layers](https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/nn/conv/simple_conv.html#SimpleConv.forward) seems to be tuples. We should have this receive `node_atributes, edge_index` or something. But I am still not sure exactly what the output should be. This will be something to find out as we go along. Therefore we can leave this as is and refactor in the future if needed

AugustoPeres (approved):

VictorGillioz:
Thanks for the feedback, I will merge this now and adapt the modules as needed while I progress. Next PR will be data loading, as there has been important changes and  I will focus on Ray right after!

AugustoPeres:
> I will focus on Ray right after!

Can I ask, how do you plan to work on ray without first having written the training loop?

VictorGillioz:
I will have a look at Ray to implement the training in a compatible manner, probably adapting the model as a Lightning module. Would that be suitable ?

AugustoPeres:
> I will have a look at Ray to implement the training in a compatible manner, probably adapting the model as a Lightning module. Would that be suitable ?

Yes. We most likely want something like:

```python
class LightningModules():
    def __init__(self, model, learning rate):
        self.model = model
```

And then integrate this using ray.

Lets sync next week to discuss the best way to do this and maybe even how to start integrating this with the `inductiva-ml-codebbase` repository

IvanPombo:
I think one could start with a Encoder-Decoder model to start integrating Ray and then move forward to integrate the processor. In this way, Victor could already develop the training loop, integrate Ray, test the framework at `inductiva-ml-codebbase` and start testing the Message Passing later with MLP instead of the identities.

AugustoPeres:
> I think one could start with a Encoder-Decoder model to start integrating Ray and then move forward to integrate the processor. In this way, Victor could already develop the training loop, integrate Ray, test the framework at `inductiva-ml-codebbase` and start testing the Message Passing later with MLP instead of the identities.

Good idea

sarmento:
It's exciting to see this moving fwd and integrating with the broader framework.

PR #8, opened 2023-05-26
Vtk loading

VictorGillioz:
I implemented two new loading methods that support meshes with multiple face shapes. They extract the nodes coordinates, the edges in the mesh and optionally the pressure at each node.

I have tested them on both .obj and .vtk files. The differences in behavior are the following:
- pyvista method allows reading of some faulty files (e.g. point_data is not the same length as points), meshio returns an error
- meshio method can return big endian arrays depending on the file it reads (observed on the .vtk files), which is not supported by torch, while pyvista automatically converts the types
- meshio method does not check for duplicated edges in its returning array. However, the data processing part transforms any edge list to a sorted, biderectional edge list with no duplicates. As a result, the two graphs are identical after processing.
- pyvista method seems to be faster than the meshio loop method when the number of points/faces becomes big

Base on these observation and code readability, I would suggest to primarely use pyvista to extract meshes from the data files.

### DIFF
--- a/meshnets/utils/data_loading.py
+++ b/meshnets/utils/data_loading.py
@@ -6,6 +6,39 @@ from typing import Tuple
 from absl import logging
 import meshio
 import numpy as np
+import pyvista as pv
+
+
+def load_edge_mesh_pv(
+    obj_path: str,
+    get_pressure: bool = True,
+    verbose: bool = False
+) -> Union[Tuple[np.ndarray, np.ndarray, np.ndarray], Tuple[np.ndarray,
+                                                            np.ndarray]]:
+    """Extract node coordinates, edges and pressure at nodes from a mesh file
+    using pyvista."""
+
+    mesh = pv.read(obj_path)
+    # `use_all_points=False` discards duplicates and renumber the nodes
+    # TODO(victor): verify that node pressure array is modified accordingly
+    # WARNING: the remeshing should be done prior on the mesh file, not here
+    edge_mesh = mesh.extract_all_edges(use_all_points=True)
+
+    nodes = edge_mesh.points
+    edge_list = edge_mesh.lines.reshape((-1, 3))[:, 1:]
+
+    if verbose:
+        logging.info('Nodes shape : %s', nodes.shape)
+        logging.info('Edge list shape : %s', edge_list.shape)
+
+    if get_pressure:
+        pressure = edge_mesh.get_array('p', preference='point')
+        if verbose:
+            logging.info('Pressure shape : %s', pressure.shape)
+
+        return nodes, edge_list, pressure
+    else:
+        return nodes, edge_list
 
 
 def load_edge_mesh_meshio(
--- a/meshnets/utils/data_loading.py
+++ b/meshnets/utils/data_loading.py
@@ -32,7 +32,8 @@ def load_edge_mesh_pv(
         logging.info('Edge list shape : %s', edge_list.shape)
 
     if get_pressure:
-        pressure = edge_mesh.get_array('p', preference='point')
+        # TODO(victor): decide if array reshape is necessary
+        pressure = edge_mesh.get_array('p', preference='point').reshape(-1, 1)
         if verbose:
             logging.info('Pressure shape : %s', pressure.shape)
 
@@ -85,7 +86,8 @@ def load_edge_mesh_meshio(
         logging.info('Edge list shape : %s', edges.shape)
 
     if get_pressure:
-        pressure = mesh.point_data['p']
+        # same behavior as mesh.points
+        pressure = mesh.point_data['p'].byteswap().newbyteorder('<')
         if verbose:
             logging.info('Pressure shape : %s', pressure.shape)
 
--- a/meshnets/utils/data_loading.py
+++ b/meshnets/utils/data_loading.py
@@ -1,4 +1,4 @@
-"""Load the mesh data"""
+""""Methods for loading mesh data"""
 
 from typing import Tuple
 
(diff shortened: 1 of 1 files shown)


VictorGillioz:
The next PR will be the adapted processing file

meshnets/utils/data_loading.py
```diff
@@ -1,17 +1,107 @@
...
 
-def load_mesh_from_obj(obj_path: str,
+    mesh = pv.read(obj_path)
+    # `use_all_points=False` discards duplicates and renumber the nodes
+    # TODO(victor): verify that node pressure array is modified accordingly
+    # WARNING: the remeshing should be done prior on the mesh file, not here
+    edge_mesh = mesh.extract_all_edges(use_all_points=True)
+
+    nodes = edge_mesh.points
+    edge_list = edge_mesh.lines.reshape((-1, 3))[:, 1:]
```
IvanPombo, line 28: This is a complex method. Can you add a comment explaining?

meshnets/utils/data_loading.py
```diff
@@ -1,17 +1,107 @@
...
+                                                            np.ndarray]]:
+    """Extract node coordinates, edges and pressure at nodes from a mesh file
+    using pyvista."""
 
-def load_mesh_from_obj(obj_path: str,
+    mesh = pv.read(obj_path)
+    # `use_all_points=False` discards duplicates and renumber the nodes
+    # TODO(victor): verify that node pressure array is modified accordingly
+    # WARNING: the remeshing should be done prior on the mesh file, not here
+    edge_mesh = mesh.extract_all_edges(use_all_points=True)
```
IvanPombo, line 25: It seems that with `use_all_points=True` also provides renumbering. Is this desired or a problem?

meshnets/utils/data_loading.py
```diff
@@ -1,17 +1,107 @@
...
+    nodes = edge_mesh.points
+    edge_list = edge_mesh.lines.reshape((-1, 3))[:, 1:]
+
+    if verbose:
+        logging.info('Nodes shape : %s', nodes.shape)
+        logging.info('Edge list shape : %s', edge_list.shape)
+
+    if get_pressure:
+        # TODO(victor): decide if array reshape is necessary
+        pressure = edge_mesh.get_array('p', preference='point').reshape(-1, 1)
```
IvanPombo, line 36: What is the objective with the reshape?

meshnets/utils/data_loading.py
```diff
@@ -1,17 +1,107 @@
...
+        if verbose:
+            logging.info('Pressure shape : %s', pressure.shape)
+
+        return nodes, edge_list, pressure
+    else:
+        return nodes, edge_list
```
IvanPombo, line 42: Your goal is to also be able to load objs with this, right?

meshnets/utils/data_loading.py
```diff
@@ -1,17 +1,107 @@
...
+    using meshio.
+    
+    Edges can include duplicates.
+    """
+
+    mesh = meshio.read(obj_path)
+    # `byteswap().newbyteorder('<')` is required because the mesh points array
+    # is sometimes in big endian byte order, which is not supported by torch
+    # Behavior observed on .vtk files
+    # TODO: this should probably be avoided at file level
```
IvanPombo, line 61: Clarify me this.

meshnets/utils/data_loading.py
```diff
@@ -1,17 +1,107 @@
...
+    nodes = mesh.points.byteswap().newbyteorder('<')
+
+    # extract all the edges in the mesh
+    edges_list = []
```
IvanPombo, line 65: Create another separate function that implements this functionality from a mesh.

IvanPombo (approved):
LGTM Victor! Left only minor comments. It is good to have both options, so that we can later decide. Pyvista seems to be more clean for now!

meshnets/utils/data_loading.py
VictorGillioz: This is unclear at the moment. I did not observe any renumbering with `use_all_points=True`, but I left a comment on the pyvista commit to ask for clarification of the method behavior. This will not be a problem for training and prediction, but if there is renumbering we risk to lose point correspondance between the original mesh and the extracted one.

meshnets/utils/data_loading.py
VictorGillioz: I added comment to the code. This is to have (nb_point, 1) instead of (nb_points,) to be coherent in the case we have more than 1 node label.

meshnets/utils/data_loading.py
VictorGillioz: Any object that does not have pressure data. It can be .obj files, it can be .vtk files with no pressure array. Moreover, the two methods should work identically on any file format accepted by meshio.

meshnets/utils/data_loading.py
VictorGillioz: How so? Should I explain what big and little endian is here

IvanPombo: Just the part of being avoided at file level. Do you mean it should be eliminated in the file to start?

meshnets/utils/data_loading.py
VictorGillioz: So I got in touch with a PyVista developer, which redirected me to VTK, and it seems that there is a small error in the pyvista docstring. So no point renumbering happens for `use_all_points=True` :)

meshnets/utils/data_loading.py
VictorGillioz: I am not completly sure but .vtk, or a least the files I have, seem to be encoded in big endian format. I don't know if and how this is avoidable.

I also realised the byteswapping was done in every case and sometimes tempered with the values, so I added a check of the endian order first.

Overall, I think this is a minor problem and the byteswapping solves it well enough not to lose time on that for now.

meshnets/utils/data_loading.py
IvanPombo: That is perfect Victor! Nicely done :)

meshnets/utils/data_loading.py
IvanPombo: Ok, I understand this now! Let's keep it simple now and maybe just use pyvista to read.

PR #9, opened 2023-05-29
Data processing

VictorGillioz:
In this PR, I implemented a new data processing method that takes as input a list of edges instead of list of faces to produce a graph object from mesh data. I also made small changes and improved the docstring of the previous processing method. 

Additionally, I changed the data visualization file so it can be called from main. I also changed the main file to include the new loading and processing methods.


meshnets/main.py
```diff
@@ -9,32 +9,37 @@
...
-    logging.info('Loading the object from %s', FLAGS.input_object)
-    nodes, cells = data_loading.load_mesh_from_obj(FLAGS.input_object,
-                                                   verbose=True)
+    #data_visualization.plot_data(FLAGS.input_object, verbose=True)
+
+    logging.info('Loading the mesh data from %s', FLAGS.input_object)
+    mesh = data_loading.load_edge_mesh_pv(FLAGS.input_object,
+                                          get_pressure=True,
+                                          verbose=False)
+    nodes, edges, pressure = mesh[0], mesh[1], mesh[2]
```
sarmento, line 27: see if this works:

nodes, edges, pressure = mesh[0:2]

meshnets/main.py
```diff
@@ -9,32 +9,37 @@
...
+    #nodes, edges, pressure = mesh[0], mesh[1], None
 
-    wind_vector = (10, 0, 0)  # dummy wind vector along the X-axis
+    wind_vector = (10, 0, 0)  # wind vector along the X-axis
```
sarmento, line 30: pls remove magic number. Use args/params

meshnets/main.py
```diff
@@ -9,32 +9,37 @@
...
-                                                   null_pressure)
+    logging.info('Building graph from the mesh data')
+    graph = data_processing.edge_mesh_to_graph(nodes, node_features, edges,
+                                               pressure)
 
     logging.info('Node features shape : %s', graph.x.shape)
     logging.info('Edge index shape : %s', graph.edge_index.shape)
     logging.info('Edge features shape : %s', graph.edge_attr.shape)
-    logging.info('Pressure label shape : %s', graph.y.shape)
+    if graph.y is not None:
```
sarmento, line 41: "if graph.y:" should be enough, right?

meshnets/main.py
```diff
@@ -9,32 +9,37 @@
 
 from meshnets.utils import data_loading
 from meshnets.utils import data_processing
+#from meshnets.utils import data_visualization
```
AugustoPeres, line 12: Remove commented code.

meshnets/main.py
```diff
@@ -9,32 +9,37 @@
...
 
 FLAGS = flags.FLAGS
-
-flags.DEFINE_string('input_object', 'data/motorBike.obj',
+flags.DEFINE_string('input_object', 'data/blob.vtk',
```
AugustoPeres, line 15: Remove the hardcoded path. Two options, either you use `os.path.join` or we can set this to `None` and mark the flag as required.

meshnets/main.py
```diff
@@ -9,32 +9,37 @@
...
                     'File path of the object to be visualized')
 
 
 def main(_):
 
-    logging.info('Loading the object from %s', FLAGS.input_object)
-    nodes, cells = data_loading.load_mesh_from_obj(FLAGS.input_object,
-                                                   verbose=True)
+    #data_visualization.plot_data(FLAGS.input_object, verbose=True)
```
AugustoPeres, line 21: remove commented code before merging to main

meshnets/main.py
```diff
@@ -9,32 +9,37 @@
...
+
+    logging.info('Loading the mesh data from %s', FLAGS.input_object)
+    mesh = data_loading.load_edge_mesh_pv(FLAGS.input_object,
+                                          get_pressure=True,
+                                          verbose=False)
+    nodes, edges, pressure = mesh[0], mesh[1], mesh[2]
+    #nodes, edges, pressure = mesh[0], mesh[1], None
```
AugustoPeres, line 28: Remove commented code

meshnets/utils/data_processing.py
```diff
@@ -28,18 +31,64 @@ def _compute_edge_attributes(data: Data, remove_pos: bool = True) -> Data:
...
+    Args shape:
+        node_coordinates: (nb_nodes, nb_space_dim)
+        node_features: (nb_nodes, nb_node_features)
+        edges: (nb_edges, 2)
+        node_labels: (nb_nodes, nb_node_labes)
+
+    The graph includes node features, edges indices, edge attributes
+    and optionally node labels."""
+
+    node_coordinates = torch.Tensor(node_coordinates)
```
AugustoPeres, line 50: It is a better practice to use `torch.tensor(node_coordinate, dtype=dtpye)`. See [this thread](https://discuss.pytorch.org/t/difference-between-torch-tensor-and-torch-tensor/30786/2)

meshnets/utils/data_processing.py
```diff
@@ -28,18 +31,64 @@ def _compute_edge_attributes(data: Data, remove_pos: bool = True) -> Data:
...
+    # Ensure a bidirectional graph with no duplicated edges
+    #pylint: disable = no-value-for-parameter
+    edge_index = to_undirected(torch.LongTensor(edges.T))
+
+    if node_labels is not None:
+        node_labels = torch.Tensor(node_labels)
+
+    graph = Data(x=node_features,
+                 edge_index=edge_index,
+                 y=node_labels,
```
AugustoPeres, line 62: Question: What are the implications of saving `y=None`?

AugustoPeres (approved):
LGTM! Thanks @VictorGillioz for a super clean pull request

meshnets/main.py
AugustoPeres: I think we should leave it like this. Suppose for example that `y` which I think is the prediction target is `0`. Then the `if` may fail when it shouldn't

IvanPombo (approved):
After the already mentioned suggestions, lgtm on my part!

meshnets/main.py
VictorGillioz: It does not. This implementation avoids raising a pylint error due to the varying size of the return from our method

meshnets/main.py
VictorGillioz: `if graph.y` is not well defined for arrays so `is not None` works better here.

meshnets/utils/data_processing.py
VictorGillioz: None is the default value so it is identical to not providing a value for y.

sarmento (approved):
LGTM!

PR #10, opened 2023-06-01
Dataset processing

VictorGillioz:
This PR includes the processing of multiple '.vtk' files to graph '.pt' files. The single file processing pipeline has been moved from main.py to a data_processing.py method. There is also a new dataset class to create a dataset from disk data that was taken from the inductiva-ml-codebase.

In the current main file, we iterate over the raw data directory, save the processed data, and then use the processed files to create torch_geometric datasets and dataloaders. 

Additionally, data loading with `get_pressure=False` now returns `None` as pressure to better integrate in the pipeline.

### DIFF
--- a/meshnets/main.py
+++ b/meshnets/main.py
@@ -5,41 +5,47 @@ and creates a Graph object from them
 """
 
 import os
+from pathlib import Path
 
 from absl import app, flags, logging
-import numpy as np
+import torch
+from torch_geometric.loader import DataLoader
 
-from meshnets.utils import data_loading
 from meshnets.utils import data_processing
+from meshnets.utils.datasets import FromDiskGraphDataset
 
 FLAGS = flags.FLAGS
-flags.DEFINE_string('input_object', os.path.join('data', 'blob.vtk'),
-                    'File path of the object to be visualized')
+flags.DEFINE_string('raw_data_dir', os.path.join('data', 'vtk'),
+                    'Path to the folder containing the raw data files')
+flags.DEFINE_string('processed_data_dir', os.path.join('data', 'pt'),
+                    'Path to the folder for the processed data files')
+flags.DEFINE_bool('process_raw', False,
+                  'Indicate if the raw data must be processed or not')
 
-# Wind vector associated with the simulation
+# Wind vector associated with the simulations
 WIND_VECTOR = (10, 0, 0)
 
 
 def main(_):
 
-    logging.info('Loading the mesh data from %s', FLAGS.input_object)
-    mesh = data_loading.load_edge_mesh_pv(FLAGS.input_object,
-                                          get_pressure=True,
-                                          verbose=False)
-    nodes, edges, pressure = mesh[0], mesh[1], mesh[2]
+    logging.info('Process the raw data : %s', FLAGS.process_raw)
+    if FLAGS.process_raw:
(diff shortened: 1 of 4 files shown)


meshnets/main.py
```diff
@@ -1,45 +1,65 @@
...
+from torch_geometric.loader import DataLoader
 
-from meshnets.utils import data_loading
 from meshnets.utils import data_processing
+from meshnets.utils.datasets import FromDiskDataset
 
 FLAGS = flags.FLAGS
-flags.DEFINE_string('input_object', os.path.join('data', 'blob.vtk'),
-                    'File path of the object to be visualized')
+flags.DEFINE_string('raw_data_dir', os.path.join('data', 'vtk'),
```
IvanPombo, line 19: What does raw mean?

IvanPombo: If it is just do differentiate between vtk and pt data, it will be best to have `vtk_data_dir` and `pt_data_dir`.

IvanPombo (approved):
LGTM!

meshnets/main.py
```diff
@@ -1,45 +1,65 @@
 """The main file to run our code
 
-Currently, it loads mesh data from a .vtk file, defines a wind vector,
-and creates a Graph object from them
+Currently, it processes mesh data to graph data and creates training
```
AugustoPeres, line 3: I think we should rename this file according to its function

meshnets/main.py
```diff
@@ -1,45 +1,65 @@
...
+        train_dataset,
+        batch_size=3,
+        shuffle=True,
+    )
+    test_dataloader = DataLoader(test_dataset, batch_size=1, shuffle=True)
+
+    for batch in train_dataloader:
+        logging.info('Training batch: %s', batch)
+    for batch in test_dataloader:
+        logging.info('Testing batch: %s', batch)
```
AugustoPeres, line 62: This part is just for testing purposes, no?

AugustoPeres (approved):
LGTM

meshnets/main.py
VictorGillioz: Yes, at this point `main.py` is a pipeline that tests all the code in sequence to assert that everything works well together.

meshnets/main.py
VictorGillioz: "raw" are the mesh files. Here they are .vtk but could also be .obj. I can rename raw to mesh if you think it's more clear

meshnets/main.py
AugustoPeres: If this file will process mesh data to graph data we could rename it to something like: ~data_preprocessing.py~ or ~mesh_to_graphs.py~

meshnets/main.py
IvanPombo: mesh is good!

meshnets/main.py
IvanPombo: test_pipeline.py?

IvanPombo (approved):

meshnets/main.py
VictorGillioz: I moved the processing loop to the processing.py file so the main method is now only a sequence of very simple steps. I think we can keep the name for now. The pipeline will certainly change again and main.py does not have a well defined role for now.

PR #11, opened 2023-06-01
Dummy model

VictorGillioz:
This PR implements a dummy MGN model, composed of an Identity encoder, Identity processor and Linear decoder. The modules have been modified to work with torch_geometric Batch produced from our dataloaders.

A pipeline for model inference is implemented in the main method. The inference on our dataset works as expected and respects the desired ouptut dimension. From there, I will implement a simple Lightning wrapper and integrate the dummy model and dataset in the MLFlow framework.

### DIFF
--- a/meshnets/main.py
+++ b/meshnets/main.py
@@ -2,6 +2,7 @@
 
 Currently, it processes mesh data to graph data and creates training
 and testing datasets and dataloaders from the processed '.pt' files.
+It then instanciates a MGN model and runs inference over the two sets.
 """
 
 import os
@@ -12,6 +13,7 @@ import torch
 from torch.utils.data import random_split
 from torch_geometric.loader import DataLoader
 
+from meshnets.modules.model import MeshGraphNet
 from meshnets.utils import data_processing
 from meshnets.utils.datasets import FromDiskDataset
 
@@ -54,12 +56,24 @@ def main(_):
         batch_size=3,
         shuffle=True,
     )
-    test_dataloader = DataLoader(test_dataset, batch_size=1, shuffle=True)
+    test_dataloader = DataLoader(test_dataset, batch_size=1, shuffle=False)
+
+    sample = dataset.get(0)
+    model = MeshGraphNet(node_features_size=sample.x.shape[1],
+                         mesh_features_size=sample.edge_attr.shape[1],
+                         output_size=sample.y.shape[1],
+                         latent_size=sample.x.shape[1],
+                         num_mlp_layers=1,
+                         message_passing_steps=1)
 
     for batch in train_dataloader:
         logging.info('Training batch: %s', batch)
+        preds = model(batch)
+        logging.info('Training batch predictions shape : %s', preds.shape)
     for batch in test_dataloader:
         logging.info('Testing batch: %s', batch)
+        preds = model(batch)
+        logging.info('Testing batch predictions shape : %s', preds.shape)
 
 
 if __name__ == '__main__':
--- a/meshnets/modules/processor.py
+++ b/meshnets/modules/processor.py
@@ -7,7 +7,7 @@ from torch_geometric.nn.conv import MessagePassing
 
 
 class MGNProcessorLayer(MessagePassing):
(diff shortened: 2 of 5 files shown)


meshnets/main.py
```diff
@@ -54,12 +56,24 @@ def main(_):
         batch_size=3,
         shuffle=True,
     )
-    test_dataloader = DataLoader(test_dataset, batch_size=1, shuffle=True)
+    test_dataloader = DataLoader(test_dataset, batch_size=1, shuffle=False)
+
+    sample = dataset.get(0)
+    model = MeshGraphNet(node_features_size=sample.x.shape[1],
```
IvanPombo, line 62: Can you describe how these features appear?

I am having a hard time figuring out and need to go through all the files. It seems that sample is a graph of the dataset (one you process above), which already contains the features  we require. I think you can add briefly on top, what the model you are constructing here entails (features of nodes and edges).

meshnets/modules/processor.py
```diff
@@ -1,10 +1,40 @@
...
+import torch
+from torch.nn import Identity
+from torch_geometric.data import Batch
+from torch_geometric.nn.conv import MessagePassing
 
-class GraphProcessor:
 
-    def __init__(self, *args):
-        pass
+class MGNProcessorLayer(MessagePassing):
```
IvanPombo, line 9: Any reason for here the MessagePassing to be of convolutional type? What is the difference from `from torch_geometric.nn import MessagePassing`?

IvanPombo (approved):
Everything is taking shape! Nice work @VictorGillioz! I only left two comments, one to detail a bit more the implementation and another out of curiosity.

Given that you are already able to infer, can one visualise the inference output?

AugustoPeres (approved):
LGTM! Left just a question related to the shapes that we want to use

meshnets/main.py
VictorGillioz: `sample` is indeed a sample of the dataset we constructed, where the dataset is composed of torch_geometric `Data` objects. What do you think I should add to clarify this?

meshnets/modules/processor.py
VictorGillioz: Both imports refer to the same class. The implement I found used `torch_geometric.nn.conv` which is a bit more verbose, but the .conv is optional.

meshnets/modules/processor.py
```diff
@@ -1,10 +1,40 @@
...
+        super().__init__()
+
+        # TODO(victor) : implement processor
+        self.processor = Identity(latent_size, message_passing_steps)
+
+    def forward(self, graph: Batch) -> Batch:
+        """Apply the GraphProcessor to a batch of graphs.
+        
+        Return the batch with processed features."""
+        return self.processor(graph)
```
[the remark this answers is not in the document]
VictorGillioz: The prediciton shape in our case is 1 feature per node. In the input, graph.x has 3 features, so I reduce that with a linear layer in the dummy decoder: (nb_nodes, 3) -> (nb_nodes, 1). We could indeed simplify the model even more by including the linear layer in the main module, or even with `return graph.x[:, 0]`.

However, I think that using the composed model structure we already have offers a good idea of what the model is going to look like at the end. This way, I was also able to check that the input/output types of each module works well with our dataset. At the end, the total model only has 4 parameters in the decoder, so it is very lightweight but should still allow us to observe some behavior during training.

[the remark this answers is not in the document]
VictorGillioz: Btw, MGNProcessor is only the Processor part of our MeshGraphNet model, and for the moment it takes in a graph batch and return a graph batch, but this might change as we go.

VictorGillioz:
Regarding the visualisation of the inference output, this is not implemented yet. I will probably focus on this part later on when we already have functional models.

meshnets/main.py
IvanPombo: I was thinking of a description of the features and their shape. But maybe I am just being picky since this file can run for different data if we change the mesh to graphs. 

Keep it like this. But in the files loading the mesh see if more documentation is needed to understand this faster.

PR #12, opened 2023-06-05
simple lightning wrapper for the MGN model

VictorGillioz:
This PR adds a simple lightning wrapper that works with the MGN model and the Batch from torch geometric.


AugustoPeres (approved):
LGTM, left just a comment

IvanPombo (approved):

PR #13, opened 2023-06-09
mlp class for encoder and decoder

VictorGillioz:
This PR adds an MLP class, representing a simple MLP defined by a list of layer widths. It offers the possibility to activate or not the last layer and to apply layer norm or not to it.

Additionally, the MGN encoder and decoder classes have been modified to include this new MLP class.

### DIFF
--- a/meshnets/modules/encoder.py
+++ b/meshnets/modules/encoder.py
@@ -1,8 +1,9 @@
 """"Define the GraphEncoder class."""
 
 import torch
-from torch.nn import Identity
 from torch_geometric.data import Batch
+
+from meshnets.modules.mlp import MLP
 
 
 class GraphEncoder(torch.nn.Module):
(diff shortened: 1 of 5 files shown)


meshnets/modules/encoder.py
```diff
@@ -18,11 +19,13 @@ def __init__(self, node_feats_size, mesh_feats_size, latent_size,
...
-        # TODO(victor) : implement encoders
-        self.node_encoder = Identity(node_feats_size, latent_size,
-                                     num_mlp_layers)
-        self.edge_encoder = Identity(mesh_feats_size, latent_size,
-                                     num_mlp_layers)
+        node_encoder_widths = [node_feats_size
+                              ] + (num_mlp_layers + 1) * [latent_size]
+        self.node_encoder = MLP(node_encoder_widths, layer_norm=True)
+
+        edge_encoder_widths = [mesh_feats_size
```
IvanPombo, line 26: mesh or edges features? Just to understand that this are local features, right?

IvanPombo (approved):
LGTM!

meshnets/modules/mlp.py
```diff
@@ -0,0 +1,48 @@
...
+
+        nb_layers = len(widths)
+
+        layers = []
+        for i in range(nb_layers - 1):
+            layers.append(Linear(widths[i], widths[i + 1]))
+            if i < (nb_layers - 2) or activate_final:
+                layers.append(ReLU())
+
+        if layer_norm:
```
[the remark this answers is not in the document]
VictorGillioz: This is over a  "classic" torch batch. In the encoder, we feed the MLP with the features from the graph batch. What happens is the following: encoder receives a graph Batch and takes the nodes features from the graph batch with shape (num_nodes, num_node_features) and feed it to the MLP.

meshnets/modules/encoder.py
VictorGillioz: You're right, edge features are probably more correct here, I changed the argument name

AugustoPeres (approved):
LGTM!

PR #14, opened 2023-06-13
improved modularity of dataset classes and num_label_features method

VictorGillioz:
This PR separate the FromDiskDataset class from torch_geometric. This makes the FromDiskClass a generic torch Dataset class for any type of data.

It then defines FromDiskGeoDataset which inherits from the torch_geometric Dataset and implements the len() and get() methods using a FromDiskDataset attribute. This class also implements a new num_label_features method following torch_geometric's implementation of similar methods.

Let me know if you think the added modularity is worth it or if I should fuse back the two classes together.

### DIFF
--- a/meshnets/utils/datasets.py
+++ b/meshnets/utils/datasets.py
@@ -1,11 +1,12 @@
-"""File containing dataset classes"""
+"""File containing dataset classes."""
 import os
 
+import numpy as np
 import torch
-from torch_geometric.data import Dataset
+from torch_geometric.data import Data, Dataset
 
 
-class FromDiskDataset(Dataset):
+class FromDiskDataset(torch.utils.data.Dataset):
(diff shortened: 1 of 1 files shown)


AugustoPeres (approved):
LGTM!!

hpenedones (commented):
Consider using "Geometric" instead of Geo which is ambiguous (geography, geological, etc). Thanks! Also, should it be even more specific than that? Is it just for "Meshes"?

VictorGillioz:
> Consider using "Geometric" instead of Geo which is ambiguous (geography, geological, etc). Thanks! Also, should it be even more specific than that? Is it just for "Meshes"?

This does not need to be more specific in my opion. This load `Data` objects representing graphs from `.pt` files in a directory. In our case, the graphs are produced from meshes but this is not a requirement.

PR #15, opened 2023-06-15
added main train file and train_model method

VictorGillioz:
PR for the model training method and the main train file

### DIFF
--- a/meshnets/train.py
+++ b/meshnets/train.py
@@ -12,9 +12,19 @@ from meshnets.utils.datasets import FromDiskGeometricDataset
 
 FLAGS = flags.FLAGS
 
+# Random seed flag
+flags.DEFINE_integer('random_seed', 21,
+                     'The seed to initialize the random number generator.')
+
 # Processed data path
 flags.DEFINE_string('processed_data_dir', os.path.join('data', 'pt'),
                     'Path to the folder for the processed data files.')
+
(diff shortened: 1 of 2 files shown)


meshnets/train.py
```diff
@@ -0,0 +1,83 @@
...
+flags.DEFINE_bool('use_gpu', False, 'Whether to use gpu or not.')
+
+# Trainer flags
+flags.DEFINE_integer('max_epochs', 10, 'The number of epochs.')
+flags.DEFINE_integer('log_every_n_steps', 1, 'How often to log within steps.')
+
+
+def main(_):
+
+    torch.manual_seed(21)
```
AugustoPeres, line 55: Instead use something like:

```python
if FLAGS.random_seed is not None:
    torch.manual_seed(21)
```

meshnets/train.py
```diff
@@ -0,0 +1,83 @@
...
+
+    dataset = FromDiskGeometricDataset(FLAGS.processed_data_dir)
+    train_dataset, validation_dataset = random_split(dataset, [0.8, 0.2])
```
AugustoPeres, line 58: Use flags for these hard-coded values

AugustoPeres (approved):
LGTM. Great job @VictorGillioz everything looks very clean!

PR #16, opened 2023-06-16
added tuning main method and in model_training

VictorGillioz:
I added a tuning main method and modified the model training method accordingly, following @AugustoPeres implementation of tuning using Ray and MLFlow.


AugustoPeres (approved):
LGTM. Everything looks very clean and I like the fact that the only repeated code between the scripts are a small set of flags and one-liners!!

PR #18, opened 2023-06-23
added MGN processor to the model

VictorGillioz:
This PR implements the MGN processor and integrates it in our model. The processor uses the torch-scatter library that currently causes problems in remote runs, so it might be changed later if a good alternative is found.

### DIFF
--- a/meshnets/modules/processor.py
+++ b/meshnets/modules/processor.py
@@ -1,22 +1,75 @@
 """"Define the Processor and ProcesserLayer classes."""
 
 import torch
-from torch.nn import Identity
+from torch.nn import Sequential
 from torch_geometric.data import Batch
 from torch_geometric.nn.conv import MessagePassing
+import torch_scatter
+
+from meshnets.modules.mlp import MLP
 
 
 class MGNProcessorLayer(MessagePassing):
-    """Single MGN Message Passing layer for a Graph object."""
+    """Single MGN Message Passing layer for graphs."""
 
-    def __init__(self, latent_size):
+    def __init__(self, latent_size, num_mlp_layers):
+        """Initialize a processor layer given the latent size
+        and number of layers in the MLPs.
+
+        Note that the latent node features and latent edge features
+        have the same size. Thus, the input of the edge processor is
+        three times the hidden dimension (one self embedding and two
+        adjacent node embeddings). The input of the node processor is
+        two times the hidden dimension (one self embedding and one sum
+        of connected edges embedings)
+        """
         super().__init__()
 
-        # TODO(victor) : implement layer
-        self.layer = Identity(latent_size)
+        edge_mlp_widths = [3 * latent_size
+                          ] + (num_mlp_layers + 1) * [latent_size]
+        self.edge_mlp = MLP(edge_mlp_widths, layer_norm=True)
+
+        node_mlp_widths = [2 * latent_size
+                          ] + (num_mlp_layers + 1) * [latent_size]
+        self.node_mlp = MLP(node_mlp_widths, layer_norm=True)
 
     def forward(self, graph: Batch) -> Batch:
-        return self.layer(graph)
+        """
+        Call the propagate method to initiate message passing and aggregation
+        on edges, and then compute the updated node features.
+        
+        It returns the graph with updated edge and node features."""
+
+        aggregated_edges, updated_edges = self.propagate(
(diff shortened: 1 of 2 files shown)


meshnets/modules/processor.py
```diff
@@ -1,22 +1,75 @@
...
+from meshnets.modules.mlp import MLP
 
 
 class MGNProcessorLayer(MessagePassing):
-    """Single MGN Message Passing layer for a Graph object."""
+    """Single MGN Message Passing layer for graphs."""
+
+    def __init__(self, latent_size, num_mlp_layers):
+        """Initialize a processor layer given the latent size
+        and number of layers in the MLPs.
```
fabiocruz, line 17: This should be a one-liner. Maybe remove the bit starting at "... given...".

meshnets/modules/processor.py
```diff
@@ -1,22 +1,75 @@
...
+
+        node_mlp_widths = [2 * latent_size
+                          ] + (num_mlp_layers + 1) * [latent_size]
+        self.node_mlp = MLP(node_mlp_widths, layer_norm=True)
 
     def forward(self, graph: Batch) -> Batch:
-        return self.layer(graph)
+        """
+        Call the propagate method to initiate message passing and aggregation
+        on edges, and then compute the updated node features.
```
fabiocruz, line 39: One-liner.

meshnets/modules/processor.py
```diff
@@ -1,22 +1,75 @@
...
+        """Initialize a processor layer given the latent size
+        and number of layers in the MLPs.
 
-    def __init__(self, latent_size):
+        Note that the latent node features and latent edge features
+        have the same size. Thus, the input of the edge processor is
+        three times the hidden dimension (one self embedding and two
+        adjacent node embeddings). The input of the node processor is
+        two times the hidden dimension (one self embedding and one sum
+        of connected edges embedings)
```
fabiocruz, line 24: It doesn't necessarily have to be the "sum of connected edges embeddings", right? It could be some other aggregation resulting in a scalar. Btw, there is a typo in "embedings".

meshnets/modules/processor.py
```diff
@@ -1,22 +1,75 @@
...
+        updated_edges = torch.cat([x_i, x_j, edge_attr], dim=1)
+        updated_edges = self.edge_mlp(updated_edges) + edge_attr
+
+        return updated_edges
+
+    def aggregate(self, updated_edges, edge_index):
+
+        # Sum the updated edges features for each node
+        _, target = edge_index
+        aggregated_edges = torch_scatter.scatter_sum(updated_edges,
```
IvanPombo, line 68: Is here the only place you use torch_scatter?

meshnets/modules/processor.py
```diff
@@ -25,13 +78,27 @@ class MGNProcessor(torch.nn.Module):
     It is a sequence of message-passing MGNProcessorLayers applied to the
     the latent node and mesh features."""
 
-    def __init__(self, latent_size, message_passing_steps):
-        """Initialize the processor given the latent size
-        and number of message passing steps"""
+    def __init__(self, latent_size, num_mlp_layers, message_passing_steps):
+        """Initialize the processor given the latent size, number of layers
+        in the MLPs and number of message passing steps."""
```
fabiocruz, line 83: One-liner.

fabiocruz (approved):
LGTM. Good job on the clear implementation of the processor @VictorGillioz! I left minor comments only.

meshnets/modules/processor.py
VictorGillioz: In this case it is a sum, but I changed it to be more general. Also fixed the typo

meshnets/modules/processor.py
VictorGillioz: Yes, this is the only operation with torch-scatter. I tried looking into torch methods but was not able to do something similar for now.

meshnets/modules/processor.py
VictorGillioz: In this case it is a sum, but I changed it to be more general. Also fixed the typo

meshnets/modules/processor.py
VictorGillioz: Yes, this is the only operation with torch-scatter. I tried looking into torch methods but was not able to do something similar for now.

PR #19, opened 2023-06-23
added main data processing file

VictorGillioz:
This PR adds a main file to run data processing from mesh files to .pt files. The loop and saving dynamic is moved from utils to this main processing file.

### DIFF
--- a/meshnets/process.py
+++ b/meshnets/process.py
@@ -26,7 +26,7 @@ flags.DEFINE_string('processed_data_dir', os.path.join('data', 'pt'),
                     'Path to the folder for the processed data files.')
 
 flags.DEFINE_bool('get_pressure', True,
-                  'Wether or not to extract pressure from the mesh files.')
+                  'Whether or not to extract pressure from the mesh files.')
 
 flags.DEFINE_bool(
     'verbose', False,
@@ -34,8 +34,10 @@ flags.DEFINE_bool(
 
 
 def main(_):
-    """Loop through a directory containing mesh files, produce a
-    graph representation of each mesh and save the graph as
+    """Process mesh data files to encoded graph data files.
+    
+    Loop through a directory containing mesh files (e.g. `.vtk`),
+    produce a graph representation of each mesh and save the graph as
     a `.pt` file in the processed data directory."""
 
     mesh_data_dir = FLAGS.mesh_data_dir
--- a/meshnets/process.py
+++ b/meshnets/process.py
@@ -1,0 +1,67 @@
+"""The main file for data processing.
+
+It processes mesh files from a given directory and produce graph data
+as '.pt' files in another directory.
+"""
+
+import os
+from pathlib import Path
+
+from absl import app
+from absl import flags
+from absl import logging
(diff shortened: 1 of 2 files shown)


meshnets/process.py
```diff
@@ -0,0 +1,67 @@
...
+
+flags.DEFINE_bool(
+    'verbose', False,
+    'Wether or not the data processing should log processing information.')
+
+
+def main(_):
+    """Loop through a directory containing mesh files, produce a
+    graph representation of each mesh and save the graph as
+    a `.pt` file in the processed data directory."""
```
fabiocruz, line 39: One liner + longer description.

meshnets/process.py
```diff
@@ -0,0 +1,67 @@
...
+FLAGS = flags.FLAGS
+
+# Path flags
+flags.DEFINE_string('mesh_data_dir', os.path.join('data', 'vtk'),
+                    'Path to the folder containing the mesh files.')
+flags.DEFINE_string('processed_data_dir', os.path.join('data', 'pt'),
+                    'Path to the folder for the processed data files.')
+
+flags.DEFINE_bool('get_pressure', True,
+                  'Wether or not to extract pressure from the mesh files.')
```
fabiocruz, line 29: Typo, "W**h**ether**.

meshnets/process.py
```diff
@@ -0,0 +1,67 @@
...
+    if not os.path.exists(processed_data_dir):
+        os.makedirs(processed_data_dir)
+
+    for mesh_file in os.listdir(mesh_data_dir):
+
+        mesh_file_path = os.path.join(mesh_data_dir, mesh_file)
+        processed_graph = data_processing.mesh_file_to_graph_data(
+            mesh_file_path,
+            WIND_VECTOR,
+            get_pressure=get_pressure,
```
fabiocruz, line 55: In which circumstance would you not want to get the pressure? Should the logic behind the data loading functions allow selecting whether or not to load velocity **and** whether or not to load pressure?

fabiocruz (approved):
LGTM! Left a question about the rationale for the data loading functions, all else is minor.

meshnets/process.py
IvanPombo: The logic with this is that I am providing the mesh with already the pressure in. Ideally, one can provide one input file with just the mesh, and another with the same mesh and pressure measurements. But as of now, I am just providing the latter on, and he can extract the input from this.

VictorGillioz: This also allows to process data for pure inference, where no pressure has been computed and is thus not present in the file.

fabiocruz: Right, so the graph with spatial coordinates as node features is always loaded and pressure is only loaded when **get_pressure=True**?

VictorGillioz: Exactly. Would you find this clearer if I change the argument to `load_pressure`?

fabiocruz: Got it. No need to rename the argument, I just wanted to understand what was going in this data loading function.

PR #20, opened 2023-06-23
updated requirements file

VictorGillioz:
Updated requirements file to install the environment dependencies using pip

### DIFF
--- a/meshnets/runtime-conda.yml
+++ b/meshnets/runtime-conda.yml
@@ -5,8 +5,11 @@ channels:
 dependencies:
   - pip:
     - absl-py
-    - torch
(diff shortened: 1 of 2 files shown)


IvanPombo (commented):
Does this already fix the problem you were having?

VictorGillioz:
The error in checks is known, and caused by the torch-scatter package. This can be solved by a sequential installation of the packages.

VictorGillioz:
> Does this already fix the problem you were having?



> Does this already fix the problem you were having?

No unfortunately not

VictorGillioz:
The latest commit fixes the requirements errors!

PR #21, opened 2023-06-27
added mlflow callbacks for checkpoint artifacts

VictorGillioz:
Added two loggers for MLFlow: The first one logs checkpoints as MLFlow artifacts each time a new checkpoint is saved. The second one logs all the checkpoints saved at the end of trainging.

### DIFF
--- a/meshnets/utils/mlf_callbacks.py
+++ b/meshnets/utils/mlf_callbacks.py
@@ -1,0 +1,43 @@
+"""Define custom lightning callbacks."""
+
+from pytorch_lightning.callbacks import ModelCheckpoint
+from pytorch_lightning.loggers.mlflow import MLFlowLogger
+
+
+class MLFlowLoggerCheckpointer(MLFlowLogger):
+    """Extend the MLFlowLogger class to log checkpoints as MLFlow artifacts
+    each time a new model is saved.
+    
+    This class saves checkpoints during training but does not delete previous
+    best checkpoints. Thus, it can lead to a high number of checkpoints
+    saved."""
+
+    def after_save_checkpoint(self,
+                              checkpoint_callback: ModelCheckpoint) -> None:
+        """ Called after model checkpoint callback saves a new checkpoint."""
+
+        self.experiment.log_artifact(self.run_id,
+                                     checkpoint_callback.best_model_path,
+                                     artifact_path='checkpoints')
+
+
+class MLFlowLoggerFinalizeCheckpointer(MLFlowLogger):
+    """Extend the MLFlowLogger class to log checkpoints as MLFlow artifacts
+    at the end of training."""
+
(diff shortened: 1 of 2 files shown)


AugustoPeres:
I have a high level comment here.

Since we are inheriting from the `mlflow logger` callback and then writing a lot of new code why not just create our own callback to log the models?

In other words, what does this gain when compared to something from scratch [like this](https://github.com/inductiva/inductiva-ml-codebase/pull/8).

Also, in the callback, I am not seeing an option to give it a directory where the models are saved. Is this an argument of the parent class? I am asking because, usually when training on a ray cluster we need all workers to have some common directory that can write to. Then this directory is given to the callbacks, this ensures that workers are saving the models to the same place and not `Path does not exist is raised`.

Therefore what happens when using these callbacks in the cluster? Can we get a print of the directory models are being saved to?


Other than this questions everything looks clean

fabiocruz:
> Since we are inheriting from the `mlflow logger` callback and then writing a lot of new code why not just create our own callback to log the models?
> 
> In other words, what does this gain when compared to something from scratch [like this](https://github.com/inductiva/inductiva-ml-codebase/pull/8).

Agreed, there seems to be repetition here...

@AugustoPeres: would it make sense to use the `MLFlowLogger` that @VictorGillioz uses here in https://github.com/inductiva/inductiva-ml-codebase/pull/8?

VictorGillioz:
In our use case, MLFlowLogger takes care of much of the needed code. It starts and stop the MLFLow run, it logs all the metrics, it also offers a method to log hyperparameters. Overall, it reduces our work to two lines in main.

The reason for the custom classes is that MLFlowLogger does not offer the possibility to save checkpoints to MLFlow (I think this is added in pytorch lightning 2 but we are using 1.6 for compatibility with ray).

However, when providing the lightning trainer with a checkpoint callback, it will call the logger's `after_save_checkpoint` method each time it saves a checkpoint. This allows us to rely on an external checkpoint callback to decide which checkpoints to save, and only take care of the artifacts logging in our custom class.

I would also argue that this does not require a lot of code, the classes are less than 20 lines.

AugustoPeres:
> However, when providing the lightning trainer with a checkpoint callback, it will call the logger's `after_save_checkpoint` 

And when is this happening? Is it every n batches? Is it at the end of each epoch?

VictorGillioz:
It is called each time the checkpoint callback saves a checkpoint. I guess it then depends on the checkpoint callback you use .

You can find the method call in the documentation [here](https://lightning.ai/docs/pytorch/1.6.0/_modules/pytorch_lightning/callbacks/model_checkpoint.html#ModelCheckpoint) at the end of the method `_save_checkpoint`. You can search for `# notify loggers`.

PR #24, opened 2023-07-10
modified dataset class for combined mesh/graph loading

VictorGillioz:
This PR implements a heavily modified dataset class. It is adapted for the new data structure on disk, where the data files are stored inside their own respective folder.

Notably, it offers the possibility to process mesh files to graph files upon class instantiation, and to save the .pt files in the same data folder.

The storage of mesh/graph data in the same folders allows to easily retrieve the mesh and graph for a single sample, even after splitting to train/validation dataset. This is particularly useful for visualisation, where we want to predict from graph files and apply the predicted field to the corresponding mesh.

### DIFF
--- a/meshnets/utils/datasets.py
+++ b/meshnets/utils/datasets.py
@@ -69,7 +69,7 @@ class FromDiskGeometricDataset(Dataset):
         This method saves each graph file in the same sample directory as
         its corresponding mesh file. The same file name is used."""
 
-        for i in range(self.len()):
+        for i in self.indices():
             mesh_file_path = self.get_mesh_path(i)
 
             processed_graph = data_processing.mesh_file_to_graph_data(
--- a/meshnets/utils/datasets.py
+++ b/meshnets/utils/datasets.py
@@ -1,86 +1,113 @@
 """File containing dataset classes."""
 import os
+import glob
+from pathlib import Path
 
 import numpy as np
 import torch
 from torch_geometric.data import Data, Dataset
 
+from meshnets.utils import data_processing
 
-class FromDiskDataset(torch.utils.data.Dataset):
-    """Reads a dataset from a given directory.
-
-    The examples are stored in `data_dir` as .pt files. The .pt
-    extension is what is commonly used to save pytorch objects such as
-    tensors, graphs, models, etc. To save an pytorch object we can use
-    ``torch.save(obj, 'path.pt')``. And then, to load it back into
-    memory we can use ``torch.load('path.pt')``.
-
-    A possible directory could look like this:
-
-    data_dir/
-        training_example_0.pt
-        training_example_1.pt
-        training_example_0.pt
-        training_example_1.pt
-
-    Note that the files do not have to be named like this.
-
-    """
-
-    def __init__(self, data_dir):
-        super().__init__()
-        self.root_dir = data_dir
-        self.files = sorted(os.listdir(data_dir))
-
-    def __len__(self):
-        return len(self.files)
-
-    def __getitem__(self, idx):
-        """The get is simply a method to retrieve an element from the
-        dataset. The only thing that exists in RAM at this point are
-        the paths to the examples. Later, this Dataset class is
-        wrapped by pytorch DataLoaders. These dataloaders will be
-        responsible for batching the data and feeding the batches to
-        the model.
-
-        Example:
-            >>> dataset = FromDiskDataset('data_dir')
-            >>> dataloader = DataLoader(dataset, batch_size=32)
-            >>> for batch in dataloader:
-            >>>     x, y = batch, batch.y
-        """
-        data_path = os.path.join(self.root_dir, self.files[idx])
-        return torch.load(data_path)
+# TODO(victor): the wind vector should be
+# obtained individually for each mesh
+WIND_VECTOR = (10, 0, 0)
 
 
 class FromDiskGeometricDataset(Dataset):
     """Reads a torch_geometric dataset from a given directory.
     
+    The data is loaded from disk each time it as accessed to reduce
+    RAM requirements. The data directory is assumed to be structured
+    as follow:
+
+    data/
+        sample_1/
+            mesh.vtk
+            graph.pt
+        sample_2/
+            mesh.vtk
+            graph.pt
+        sample_3/
+            mesh.vtk
+            graph.pt
+
+    Note that the directory, file names, and file extensions can be different.
+    `process_data` allows to produce the graph files upon instanciating the 
+    dataset class from a folder containing only the mesh files.
+
     This class inherits from torch_geometric Dataset and implements its
-    abstract methods `len` and `get` from the FromDiskDataset class.
-    This offers acces to the torch_geometric Dataset attributes when the .pt
-    files retrieved from disk are torch_geometric Data.
+    abstract methods `len` and `get`. This offers acces to the torch_geometric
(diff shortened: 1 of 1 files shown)


IvanPombo (approved):
Lgtm!

AugustoPeres (approved):

meshnets/utils/datasets.py
```diff
@@ -1,86 +1,113 @@
...
-        return torch.load(data_path)
+# TODO(victor): the wind vector should be
+# obtained individually for each mesh
+WIND_VECTOR = (10, 0, 0)
 
 
 class FromDiskGeometricDataset(Dataset):
     """Reads a torch_geometric dataset from a given directory.
     
+    The data is loaded from disk each time it as accessed to reduce
```
fabiocruz, line 20: Typo, "...it *is* accessed...*

meshnets/utils/datasets.py
```diff
@@ -1,86 +1,113 @@
...
+            graph.pt
+        sample_2/
+            mesh.vtk
+            graph.pt
+        sample_3/
+            mesh.vtk
+            graph.pt
+
+    Note that the directory, file names, and file extensions can be different.
+    `process_data` allows to produce the graph files upon instanciating the
```
fabiocruz, line 36: Typo, "instantiating"

meshnets/utils/datasets.py
```diff
@@ -1,86 +1,113 @@
...
+
+        self.mesh_file_ext = mesh_file_ext
+        self.graph_file_ext = graph_file_ext
+
+        if process_data:
+            self.process_data()
+
+    def process_data(self) -> None:
+        """Process the mesh files in the data directory to their graph
+        representation. Save the graph file in the same sample directory."""
```
fabiocruz, line 69: Write this up in a one-liner + other lines.

meshnets/utils/datasets.py
```diff
@@ -1,86 +1,113 @@
...
+                 *args,
+                 process_data: bool = False,
+                 mesh_file_ext: str = '.vtk',
+                 graph_file_ext: str = '.pt',
+                 **kwargs) -> None:
         super().__init__(*args, **kwargs)
-        self.from_disk_dataset = FromDiskDataset(data_dir=data_dir)
+
+        self.root_dir = data_dir
+        self.samples = sorted(os.listdir(data_dir))
```
fabiocruz, line 59: I'd count only subdirs of `data_dir` as samples. You may want to have a dataset summary file, e.g. `data_summary.txt` in `data_dir`.

meshnets/utils/datasets.py
```diff
@@ -1,86 +1,113 @@
...
+
+    def get_sample_path(self, idx: int) -> str:
+        """Get the path to the sample directory at the given index."""
+        return os.path.join(self.root_dir, self.samples[idx])
+
+    def get_mesh_path(self, idx: int) -> str:
+        """Get the path to the mesh file at the given index."""
+        sample_path = self.get_sample_path(idx)
+        mesh_path = glob.glob(
+            os.path.join(sample_path, f'*{self.mesh_file_ext}'))[0]
```
fabiocruz, line 93: This will not work if you have other vtk files in `sample_path`, which is very likely... Should we set the expected names of the vtk files?

meshnets/utils/datasets.py
```diff
@@ -1,86 +1,113 @@
...
+        return mesh_path
+
+    def get_graph_path(self, idx: int) -> str:
+        """Get the path to the graph file at the given index."""
+        sample_path = self.get_sample_path(idx)
+        graph_path = glob.glob(
+            os.path.join(sample_path, f'*{self.graph_file_ext}'))[0]
```
fabiocruz, line 100: Can't you use `Path(...).with_suffix(...)` here too? It would be more readable.

meshnets/utils/datasets.py
```diff
@@ -1,86 +1,113 @@
...
+        self.mesh_file_ext = mesh_file_ext
+        self.graph_file_ext = graph_file_ext
+
+        if process_data:
+            self.process_data()
+
+    def process_data(self) -> None:
+        """Process the mesh files in the data directory to their graph
+        representation. Save the graph file in the same sample directory."""
+        for i in range(self.len()):
```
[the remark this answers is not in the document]
VictorGillioz: To make use of the get_mesh_path. I think it improves modularity in this case.

The alternative would be to list the dirs that are already stored in self.samples and then compose the path, which is already taken care of by the get_mesh_path method.

meshnets/utils/datasets.py
```diff
@@ -1,86 +1,113 @@
...
     It also offer a new property called 'num_label_features' following
     the implementation of 'num_node_features' and 'num_edge_features'
     in torch_geometric Dataset.
     """
 
-    def __init__(self, data_dir, *args, **kwargs):
+    def __init__(self,
+                 data_dir: str,
+                 *args,
+                 process_data: bool = False,
```
fabiocruz, line 52: I think it would be better to leave this functionality for a separate method of the class to decouple initialization from processing, e.g. `FromDiskGeometricDataset.convert_mesh_to_graph_data()`.

fabiocruz (approved):
Thanks @VictorGillioz. I'm approving it, but I think there are a couple of things that should be improved. Let me know your thoughts about my comments below.

meshnets/utils/datasets.py
VictorGillioz: I access here the first vtk file found in the sample dir, if there are multiple the behavior it will probably access a random one. 

What do you suggest? I don't know what why there would be multiple .vtk files in a single sample dir.

meshnets/utils/datasets.py
VictorGillioz: Path(...).with_suffix(...) is used to change the suffix. In our case it is used when creating the .pt file in order to keep the same path and file name but change the file extension.

What I could replace this with is os.listdir() and then list comprehension to get the files with .pt extension. Would you prefer that?

meshnets/utils/datasets.py
VictorGillioz: Sure, I can do that

meshnets/utils/datasets.py
VictorGillioz: That could be an option yes, I can easily replace self.mesh_file_ext by self.mesh_file_name, and similarly for the graph file name. This would allow multiple .vtk and .pt files for the same sample, allowing also for different graph representation in the same data dir.

This comes at the cost of having to define the file names but it might be worth it. Let me know if I should go for this option.

fabiocruz: I'd go with this option, yes. Thanks @VictorGillioz!

meshnets/utils/datasets.py
fabiocruz: If you set file names for both mesh and graph files, this would be trivial to build and this discussion isn't needed anymore, right?

VictorGillioz: Indeed :)

meshnets/utils/datasets.py
VictorGillioz: I added a warning when the path to a file that does not exist is returned. I can also use assert, throw an error, or do nothing. Do you know best practice in this case?

VictorGillioz:
I added a remove files method, so for example .pt files can be easily removed if we don't want them anymore in the dataset.

PR #25, opened 2023-07-11
Data norm

VictorGillioz:
PR that implements data normalization.

Stats are computable inside the Dataset class. They are accessed and logged to mlflow in the model_training method. The normalization occurs in the lightning wrapper.

### DIFF
--- a/meshnets/modules/lightning_wrapper.py
+++ b/meshnets/modules/lightning_wrapper.py
@@ -60,7 +60,7 @@ class MGNLightningWrapper(pl.LightningModule):
         return batch
 
     def forward(self, batch: Batch) -> torch.Tensor:
-        # Normalize the batch according to the given metrics before
+        """Normalize the batch before making a forward pass in the model."""
         batch = self.normalize_input(batch)
         return self.model(batch)
 
--- a/meshnets/modules/lightning_wrapper.py
+++ b/meshnets/modules/lightning_wrapper.py
@@ -12,52 +12,69 @@ class MGNLightningWrapper(pl.LightningModule):
                  model: torch.nn.Module,
                  data_stats: dict,
                  learning_rate: float = 1e-3):
+        """Initialize for a given model, a dictionnary of
+        data statistics for normalization, and a learning rate.
+        
+        The statistics dictionnary is expected to have
+        the following fields:
+            'x_mean'
+            'x_std'
+            'edge_attr_mean'
+            'eadge_attr_std'
+            'y_mean'
+            'y_std'
+        """
         super().__init__()
+
         self.model = model
         self.data_stats = data_stats
         self.learning_rate = learning_rate
 
-    def normalize(self, batch: Batch) -> Batch:
+    def normalize_input(self, batch: Batch) -> Batch:
+        """Normalize the input features in a batch.
+        
+        The node and edge features are normalized using the mean and std
+        statistics found in the data_stats dictionnary."""
 
-        # Normalize the input features
-        batch.x = (batch.x -
-                   self.data_stats['x_mean']) / self.data_stats['x_std']
+        # Send the stats to the same device as the data and normalize
+        batch.x = (batch.x - self.data_stats['x_mean'].to(
+            batch.x.device)) / self.data_stats['x_std'].to(batch.x.device)
 
-        batch.edge_attr = (batch.edge_attr - self.data_stats['edge_attr_mean']
-                          ) / self.data_stats['edge_attr_std']
+        batch.edge_attr = (
+            batch.edge_attr - self.data_stats['edge_attr_mean'].to(
+                batch.edge_attr.device)) / self.data_stats['edge_attr_std'].to(
+                    batch.edge_attr.device)
 
-        # Normalize the labels
-        if batch.y is not None:
-            batch.y = (batch.y -
-                       self.data_stats['y_mean']) / self.data_stats['y_std']
+        return batch
+
+    def normalize_labels(self, batch: Batch) -> Batch:
+        """Normalize the labels in a batch.
+        
+        The labels are normalized using the mean and std statistics
+        found in the data_stats dictionnary."""
+
+        # Send the stats to the same device as the data and normalize
+        batch.y = (batch.y - self.data_stats['y_mean'].to(
+            batch.y.device)) / self.data_stats['y_std'].to(batch.y.device)
 
         return batch
 
     def forward(self, batch: Batch) -> torch.Tensor:
-        batch = self.normalize(batch)
+        # Normalize the batch according to the given metrics before
+        batch = self.normalize_input(batch)
         return self.model(batch)
 
     def training_step(self, batch: Batch, _) -> dict:
-
-        batch = self.normalize(batch)
-
         loss = self.compute_loss(batch)
         # batch_size can vary depending on the number of nodes in the graphs
         # so we pass it explicitly to the logger
-        # TODO(victor): check that batch_size is set correctly or if it
-        # should be the number of graphs in the batch
         self.log('loss', loss, on_epoch=True, batch_size=batch.num_nodes)
         return {'loss': loss}
 
     def validation_step(self, batch: Batch, _) -> dict:
-
-        batch = self.normalize(batch)
-
         val_loss = self.compute_loss(batch)
         # batch_size can vary depending on the number of nodes in the graphs
         # so we pass it explicitly to the logger
-        # TODO(victor): check that batch_size is set correctly or if it
-        # should be the number of graphs in the batch
         self.log('val_loss',
                  val_loss,
                  on_epoch=True,
@@ -66,7 +83,8 @@ class MGNLightningWrapper(pl.LightningModule):
         return {'val_loss': val_loss}
 
     def compute_loss(self, batch: Batch) -> torch.Tensor:
-        predictions = self.model(batch)
+        predictions = self.forward(batch)
+        batch = self.normalize_labels(batch)
         loss = torch.nn.functional.mse_loss(predictions, batch.y)
         return loss
 
--- a/meshnets/modules/lightning_wrapper.py
+++ b/meshnets/modules/lightning_wrapper.py
@@ -8,12 +8,39 @@ import pytorch_lightning as pl
 class MGNLightningWrapper(pl.LightningModule):
     """Simple pytorch_lightning wrapper for the MeshGraphNet training logic."""
 
-    def __init__(self, model: torch.nn.Module, learning_rate: float = 1e-3):
+    def __init__(self,
+                 model: torch.nn.Module,
+                 data_stats: dict,
+                 learning_rate: float = 1e-3):
         super().__init__()
         self.model = model
+        self.data_stats = data_stats
         self.learning_rate = learning_rate
 
+    def normalize(self, batch: Batch) -> Batch:
+
+        # Normalize the input features
+        batch.x = (batch.x -
+                   self.data_stats['x_mean']) / self.data_stats['x_std']
+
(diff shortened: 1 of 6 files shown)


meshnets/modules/lightning_wrapper.py
```diff
@@ -23,6 +50,9 @@ def training_step(self, batch: Batch, _) -> dict:
         return {'loss': loss}
 
     def validation_step(self, batch: Batch, _) -> dict:
+
+        batch = self.normalize(batch)
```
IvanPombo, line 54: Why not avoid doing this again, by using the self.forward inside the loss computation?

IvanPombo: Same above.

meshnets/utils/datasets.py
```diff
@@ -1,90 +1,197 @@
...
+        The only things that exist in RAM at this point are the paths 
+        to the samples. This method loads and returns the sample from disk."""
+
+        graph_path = self.get_graph_path(idx)
+        return torch.load(graph_path)
+
+    def get_stats(self):
+        """Compute and return the dataset statistics."""
+
+        sample = self[0]
```
IvanPombo, line 135: This is very cryptic.

meshnets/utils/datasets.py
```diff
@@ -1,90 +1,197 @@
...
+        edge_attr_mean = torch.zeros(sample.edge_attr.shape[1:])
+        edge_attr_std = torch.zeros(sample.edge_attr.shape[1:])
+        edge_attr_num = 0
+
+        y_mean = torch.zeros(sample.y.shape[1:])
+        y_std = torch.zeros(sample.y.shape[1:])
+        y_num = 0
+
+        # Iterate over the samples in the dataset to compute the means
+        for sample in self:
```
IvanPombo, line 151: Does this actually provide the samples?

IvanPombo (approved):
Minor comments, lgtm!

meshnets/utils/datasets.py
```diff
@@ -1,90 +1,197 @@
...
 import torch
 from torch_geometric.data import Data, Dataset
 
+from meshnets.utils import data_processing
 
-class FromDiskDataset(torch.utils.data.Dataset):
-    """Reads a dataset from a given directory.
+# TODO(victor): the wind vector should be
+# obtained individually for each mesh
+WIND_VECTOR = (10, 0, 0)
```
fabiocruz, line 13: I actually don't think we should use the wind vector as node features. It is adding unnecessary calculations since the wind vector is the same for all nodes. It should either be a global feature of the graph, or at this stage (with one wind velocity vector only) none at all. Let's open an issue to discuss this further.

meshnets/utils/datasets.py
```diff
@@ -1,90 +1,197 @@
...
-        This is the data object at index :obj:`idx` in the dataset from disk."""
-        return self.from_disk_dataset[idx]
+        The only things that exist in RAM at this point are the paths 
+        to the samples. This method loads and returns the sample from disk."""
+
+        graph_path = self.get_graph_path(idx)
+        return torch.load(graph_path)
+
+    def get_stats(self):
+        """Compute and return the dataset statistics."""
```
fabiocruz, line 133: Which statistics. Give a few more details.

meshnets/utils/datasets.py
```diff
@@ -1,90 +1,197 @@
...
+        edge_attr_num = 0
+
+        y_mean = torch.zeros(sample.y.shape[1:])
+        y_std = torch.zeros(sample.y.shape[1:])
+        y_num = 0
+
+        # Iterate over the samples in the dataset to compute the means
+        for sample in self:
+
+            x_mean += torch.sum(sample.x, dim=0)
```
fabiocruz, line 153: Wouldn't something like this work?
```python
x = torch.tensor([sample.x for sample in self])
x_mean = torch.mean(x, dim=0)
x_std = torch.std(x, dim=0)
```

meshnets/utils/model_training.py
```diff
@@ -60,6 +60,9 @@ def train_model(config, experiment_config, train_dataset, validation_dataset):
     # Checkpoint config
     save_top_k = experiment_config['save_top_k']
 
+    # Compute the training dataset stats for normalization
+    train_stats = train_dataset.dataset[train_dataset.indices].get_stats()
```
fabiocruz, line 64: Why doesn't `train_dataset.get_stats()` work?

fabiocruz (approved):
Looking good, @VictorGillioz! Thank you! I left some questions/suggestions to simplify the dataset method to get stats.

fabiocruz:
> The normalization occurs in the lightning wrapper.

Calling @AugustoPeres to chime in on whether this is a good practice.

meshnets/utils/datasets.py
VictorGillioz: This extracts the first dataset sample to obtain the data shapes

meshnets/utils/datasets.py
VictorGillioz: Yes, it iterates over all the samples in the dataset we are in. Additionally, if we take a slice of the dataset, it will iterate over the samples in the slice.

meshnets/utils/datasets.py
```diff
@@ -1,90 +1,197 @@
...
 
-    A possible directory could look like this:
-
-    data_dir/
-        training_example_0.pt
-        training_example_1.pt
-        training_example_0.pt
-        training_example_1.pt
+class FromDiskGeometricDataset(Dataset):
+    """Reads a torch_geometric dataset from a given directory.
```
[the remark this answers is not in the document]
VictorGillioz: I don't know, I messed up the branches I guess..

meshnets/utils/datasets.py
VictorGillioz: Right, but for now I guess we can see that as a placeholder that we can later change when augmenting the model dynamics.

meshnets/utils/datasets.py
VictorGillioz: I think it would, but it would go against the goal of reading the data from disk, as it would have to load and keep everything in memory at the same time.

meshnets/utils/model_training.py
VictorGillioz: I think the Subset class of torch does not offer access to their hyperclass methods such as .get_stats(), but I will check that.

meshnets/modules/lightning_wrapper.py
```diff
@@ -8,12 +8,39 @@
...
         self.learning_rate = learning_rate
 
+    def normalize(self, batch: Batch) -> Batch:
+
+        # Normalize the input features
+        batch.x = (batch.x -
+                   self.data_stats['x_mean']) / self.data_stats['x_std']
+
+        batch.edge_attr = (batch.edge_attr - self.data_stats['edge_attr_mean']
+                          ) / self.data_stats['edge_attr_std']
```
AugustoPeres, line 27: Isn't this standard `Batch Normalization`. Could we not replace this by a batch normalization layer. This has an implementation for [torch geometric](https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/nn/norm/batch_norm.html).

Also, I think it should be:

```
batch.x = (batch.x - self.data_stats['x_mean']) / (self.data_stats['x_std'] + epsilon)
```

where `epsilon` is a small constant for added numerical stability

meshnets/utils/datasets.py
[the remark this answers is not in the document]
VictorGillioz: Fixed!

meshnets/modules/lightning_wrapper.py
AugustoPeres: Additionally batch normalization is difference during training and inference. From the [original paper](https://arxiv.org/pdf/1502.03167.pdf).

*The normalization of activations that
depends on the mini-batch allows efficient training, but is
neither necessary nor desirable during inference; we want
the output to depend only on the input, deterministically.
For this, once the network has been trained, we use the
normalization*

And here I am not seeing difference in the inference step which can cause high performance models to look poorly trained.

The default batch normalization layers do this automatically when we call `model.eval()`

AugustoPeres:
> > The normalization occurs in the lightning wrapper.
> 
> Calling @AugustoPeres to chime in on whether this is a good practice.

I am unsure on what the goal here was. But either way:

1. I f the goal is to simply normalize the data in the dataset then I think that this should be done in the dataset class to keep lightning wrapper seperate from the normalization logic. Especially since we are already computing the mean and standard deviations there I see no reason not to normalize on either: `get_item` or even on the `init`;
2. If the goal was to implement batch normalization I think that this should be done using `BatchNorm` default layers for the reasons pointed out [here](https://github.com/inductiva/meshnets/pull/25#discussion_r1260908218)

meshnets/utils/datasets.py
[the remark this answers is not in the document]
fabiocruz: Thank you! 🙏

meshnets/utils/datasets.py
fabiocruz: Agreed. Let's just open the issue so we don't forget that the wind vector is being used as a node feature, equal for all nodes.

meshnets/utils/datasets.py
fabiocruz: Right, ok.

VictorGillioz:
The goal is to be able to train a model on a dataset with specific statistics, and later do inference on new samples. For this we would need to store the statistics used in the model somewhere so we can use them back later.

AugustoPeres (approved):
LGTM! I am still not sure if normalizing in the `lightning wrapper` is the best choice so far but we can figure this out as we go

VictorGillioz:
Not sure either, the implementations I found do that in the model itself I think. Of course we can still adapt that at some point.

meshnets/modules/lightning_wrapper.py
VictorGillioz: No real reason, but I guess I can indeed use the lightning module forward method instead yes.

meshnets/modules/lightning_wrapper.py
VictorGillioz: This is not really like batch normalization. What we do here is normalise all the data with respect to the training set statistics. This way, the training set features and labels will have 0 mean and 1 variance, it does not apply at batch level. 

We then have to normalise the test set in the same way otherwise the features between train and test will have completely different ranges.

meshnets/modules/lightning_wrapper.py
```diff
@@ -8,26 +8,73 @@
...
+    def normalize_input(self, batch: Batch) -> Batch:
+        """Normalize the input features in a batch.
+        
+        The node and edge features are normalized using the mean and std
+        statistics found in the data_stats dictionnary."""
+
+        # Send the stats to the same device as the data and normalize
+        batch.x = (batch.x - self.data_stats['x_mean'].to(
+            batch.x.device)) / self.data_stats['x_std'].to(batch.x.device)
+
```
AugustoPeres, line 42: Move to the GPU on __init__ in order to perform this just once

VictorGillioz: I think this implementation is the better option:
- at init time, we do not yet have information about where lightning is going to put the data (cpu/gpu) so we would need to 'hardcode' that using the config parameters
- if we train on multiple GPUs (e.g. tuning), I believe we would need to send the data to the same GPU as the batch during training, not any GPU

I timed the operation and it takes 2.5s for 100'000 repetitions, which I think is negligible.

AugustoPeres: Good points

PR #27, opened 2023-07-13
added multiple debugging callbacks

VictorGillioz:
This PR renames the callback file where the MLFlow loggers were located.

It adds three custom callbacks :
- gpu usage logger
- model gradient logger
- batch size logger


meshnets/utils/callbacks.py
```diff
@@ -0,0 +1,125 @@
...
+            self._checkpoint_callback = checkpoint_callback
+
+    def finalize(self, status: str) -> None:
+        """Log the checkpoints as MLFlow artifacts at the end of training."""
+
+        if self._checkpoint_callback is not None:
+            self.experiment.log_artifacts(self.run_id,
+                                          self._checkpoint_callback.dirpath,
+                                          artifact_path='checkpoints')
+
```
AugustoPeres, line 49: Didn't we have these callbacks before. Why are they in green?

meshnets/utils/callbacks.py
```diff
@@ -0,0 +1,125 @@
...
+
+class GPUUsage(Callback):
```
AugustoPeres, line 51: Why no just install the `inductiva-ml-codebase` and import from there?

meshnets/utils/callbacks.py
```diff
@@ -0,0 +1,125 @@
...
+    """Logs the number of nodes and edges in the current
+    torch geometric batch."""
+
+    def on_train_batch_start(self,
+                             trainer,
+                             pl_module,
+                             batch: Batch,
+                             batch_idx: int,
+                             unused: int = 0) -> None:
+        trainer.logger.log_metrics(
```
AugustoPeres, line 120: Is this `trainer.logger` going to mlflow? If so this will probably be a very large bottleneck when logging to an external server. Maybe we could have some `num_nodes_history` and then every `k` steps log everything in batch.

meshnets/utils/callbacks.py
```diff
@@ -0,0 +1,125 @@
...
+        if (trainer.global_step +
+                1) % self.log_freq == 0 or trainer.is_last_batch:
+            for param in trainer.model.parameters():
+                total_norm = 0
+                if param.grad is not None:
+                    param_norm = param.grad.data.norm(2)
+                    total_norm += param_norm.item()**2
+
+            gradient_norm = total_norm**0.5
+            trainer.logger.log_metrics({'gradient_norm': gradient_norm},
```
AugustoPeres, line 106: Is this trainer.logger going to mlflow? If so this will probably be a very large bottleneck when logging to an external server. Maybe we could have some num_nodes_history and then every k steps log everything in batch.

(Same comment below)

meshnets/utils/callbacks.py
```diff
@@ -0,0 +1,125 @@
...
+                                       step=trainer.global_step)
+
+
+class GeometricBatchSize(Callback):
+    """Logs the number of nodes and edges in the current
+    torch geometric batch."""
```
fabiocruz, line 112: One-liner.

meshnets/utils/callbacks.py
fabiocruz: I guess it's because @VictorGillioz renamed the callbacks file **and** added content to it.

fabiocruz (approved):
LGTM. I identified minor stuff only, but I think @AugustoPeres has important points about reusing callbacks from the ML infra repo and bout the logs to MLflow that should be sorted out before this is merged.

meshnets/utils/callbacks.py
VictorGillioz: Indeed, sorry about that :)

The two first callbacks are identical.

meshnets/utils/callbacks.py
VictorGillioz: This is a good idea, I am currently looking into adding a buffer to the custom mlflow logger.

meshnets/utils/callbacks.py
VictorGillioz: I created new custom loggers that are not in the ml-codebase, and this implementation worked well with the logger I am using. I also made small modifications to the GPUUsage class from the ml-codebase

meshnets/utils/callbacks.py
VictorGillioz: This is a good idea, I am currently looking into adding a buffer to the custom mlflow logger.

PR #29, opened 2023-07-18
load model method

VictorGillioz:
PR to add a method that allows the loading of a model from mlflow

### DIFF
--- a/meshnets/utils/model_loading.py
+++ b/meshnets/utils/model_loading.py
@@ -13,12 +13,44 @@ from meshnets.modules.lightning_wrapper import MGNLightningWrapper
 from meshnets.utils.datasets import FromDiskGeometricDataset
 
 
-def load_model_mgn(tracking_uri: str, run_id: str, ckpt: Union[int, str],
-                   dataset: FromDiskGeometricDataset) -> MGNLightningWrapper:
-    """Load a model from an MLFlow run and a given checkpoints.
+def load_model_from_mlflow(tracking_uri: str, run_id: str,
+                           ckpt: Union[int, str]) -> MGNLightningWrapper:
+    """Load a model from an MLFlow run and a given checkpoint.
     
     The checkpoint can be specified as either an index or the name of the
-    checkpoint."""
+    checkpoint.
+    
+    This method requires the hyperparameters to be saved in the checkpoint."""
+
+    client = mlflow.tracking.MlflowClient(tracking_uri=tracking_uri)
+
+    with tempfile.TemporaryDirectory() as temp_dir:
+        if isinstance(ckpt, int):
+            artifacts = client.list_artifacts(run_id, 'checkpoints')
+            ckpt_path = client.download_artifacts(run_id,
+                                                  path=artifacts[ckpt].path,
+                                                  dst_path=temp_dir)
+        elif isinstance(ckpt, str):
+            ckpt_path = client.download_artifacts(run_id,
+                                                  path=os.path.join(
+                                                      'checkpoints', ckpt),
+                                                  dst_path=temp_dir)
+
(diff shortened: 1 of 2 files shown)


meshnets/utils/model_loading.py
```diff
@@ -0,0 +1,51 @@
+"""Methods for loading models"""
+
+import os
+import tempfile
+from typing import Union
+
+import mlflow
+# pylint: disable=unused-import
```
AugustoPeres, line 8: Why disable this warning instead of removing the import

AugustoPeres (approved):

meshnets/utils/model_loading.py
```diff
@@ -0,0 +1,51 @@
...
+from torch import tensor  # used to eval tensors
+
+from meshnets.modules.model import MeshGraphNet
+from meshnets.modules.lightning_wrapper import MGNLightningWrapper
+from meshnets.utils.datasets import FromDiskGeometricDataset
+
+
+def load_model_mgn(tracking_uri: str, run_id: str, ckpt: Union[int, str],
+                   dataset: FromDiskGeometricDataset) -> MGNLightningWrapper:
```
fabiocruz, line 17: To make it explicit that this is a function to load a model from MLflow, I'd name it `load_mgn_model_from_mlfow(...)`.

meshnets/utils/model_loading.py
```diff
@@ -0,0 +1,51 @@
...
+    
+    The checkpoint can be specified as either an index or the name of the
+    checkpoint."""
+
+    client = mlflow.tracking.MlflowClient(tracking_uri=tracking_uri)
+    run = client.get_run(run_id)
+
+    params = run.data.params
+    # pylint: disable=eval-used
+    params = {k: eval(v) for k, v in params.items()}
```
fabiocruz, line 28: Is there an alternative to using `eval`?

fabiocruz (approved):

meshnets/utils/model_loading.py
VictorGillioz: Done!

meshnets/utils/model_loading.py
VictorGillioz: The import is used implicitly by the `eval` method to evaluate the parameters `tensor([...])

meshnets/utils/model_loading.py
VictorGillioz: Because the parameters are read as strings I am not sure, but this method is now deprecated

meshnets/utils/model_loading.py
```diff
@@ -13,12 +13,44 @@
 from meshnets.utils.datasets import FromDiskGeometricDataset
 
 
-def load_model_mgn(tracking_uri: str, run_id: str, ckpt: Union[int, str],
-                   dataset: FromDiskGeometricDataset) -> MGNLightningWrapper:
-    """Load a model from an MLFlow run and a given checkpoints.
+def load_model_from_mlflow(tracking_uri: str, run_id: str,
+                           ckpt: Union[int, str]) -> MGNLightningWrapper:
```
IvanPombo, line 17: checkpoint?

VictorGillioz:
I added the save_hyperparameters() call to the lightning wrapper, this removes the necessity to instantiate the model from the mlflow parameters because it is saved in the checkpoint. This results in an increased checkpoint file size (e.g. 160Kb to 200Kb for a small model) but the loading is greatly simplified.

IvanPombo (approved):

meshnets/utils/model_loading.py
VictorGillioz: Done

PR #30, opened 2023-08-04
modified data processing and flag

VictorGillioz:
Modified the main processing file to make use of our new dataset implementation.

The processing file now consist only of the two following steps:
- instantiate a dataset class given the data directory
- use the built in processing method of our dataset class to create the processed .pt files


IvanPombo (approved):
Top!! But with this refactor, I think you could provide a better description in the PR of what you are doing.

PR #32, opened 2023-08-08
Visualization

VictorGillioz:
PR for three visualization methods for meshes and array field

### DIFF
--- a/meshnets/utils/data_visualization.py
+++ b/meshnets/utils/data_visualization.py
@@ -36,8 +36,8 @@ def plot_mesh_with_scalars(mesh_path: str,
     mesh = mesh.rotate_z(rot_z)
 
     if scalars is None:
-        warnings.warn("Argument 'scalars' is None. Displaying with no values.")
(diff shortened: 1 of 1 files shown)


meshnets/utils/data_visualization.py
```diff
@@ -1,16 +1,96 @@
...
+    mesh.plot(scalars=scalars,
+              show_edges=True,
+              cmap='RdBu_r',
+              clim=clim,
+              off_screen=off_screen,
+              screenshot=screenshot)
+
+
+def plot_mesh_comparison(mesh_path: str,
+                         scalars_1: np.array,
```
sarmento, line 51: if this is the groundtruth, then call it so. Scalar_1 is not a very good name :)
The same for prediction.

sarmento (approved):
LGTM.

In general, if you submit a pice of code that does visualizations, try to also add images to the PR on git hub.

VictorGillioz:
plot of a mesh
plot comparison

PR #33, opened 2023-08-10
added non-mean gpu usage callback

VictorGillioz:
Added a GPU Usage callback that only access usage at the logging step. This samples the gpu usage instead of computing a average like the previous one

### DIFF
--- a/meshnets/utils/callbacks.py
+++ b/meshnets/utils/callbacks.py
@@ -62,9 +62,35 @@ class GPUUsage(Callback):
     def on_batch_end(self, trainer, pl_module):
         if (trainer.global_step +
                 1) % self.log_freq == 0 or trainer.is_last_batch:
+            try:
+                all_gpus = GPUtil.getGPUs()
+                for gpu in all_gpus:
+                    trainer.logger.log_metrics(
+                        {f'{gpu.uuid}_memory_usage': gpu.memoryUtil},
+                        step=trainer.global_step)
+            # pylint: disable=broad-except
+            except Exception as e:
+                warnings.warn('Something went wrong fetching GPU information.'\
+                              f'Failed with exception {e}')
+
+
+class GPUUsageMean(Callback):
+    """Track and log GPU usage."""
+
+    def __init__(self, log_freq=50):
+        """
+        Args:
+            log_freq: The frequency, in batches, at which to log GPU usage.
+        """
+        self.log_freq = log_freq
+        self.gpu_usage_history = {}
+
+    def on_batch_end(self, trainer, pl_module):
+        if (trainer.global_step +
+                1) % self.log_freq == 0 or trainer.is_last_batch:
             for gpu_uuid, gpu_usage in self.gpu_usage_history.items():
                 trainer.logger.log_metrics(
-                    {f'gpu_memory_{gpu_uuid}_usage': np.mean(gpu_usage)},
+                    {f'{gpu_uuid}_mean_memory_usage': np.mean(gpu_usage)},
                     step=trainer.global_step)
                 self.gpu_usage_history = {}
         self._update_gpu_usage()
@@ -79,7 +105,7 @@ class GPUUsage(Callback):
                 self.gpu_usage_history[uuid].append(gpu.memoryUtil)
         # pylint: disable=broad-except
         except Exception as e:
-            warnings.warn('Something went wrong fetching GPU information.\n'\
+            warnings.warn('Something went wrong fetching GPU information.'\
                           f'Failed with exception {e}')
 
 
--- a/meshnets/utils/callbacks.py
+++ b/meshnets/utils/callbacks.py
@@ -49,7 +49,7 @@ class MLFlowLoggerFinalizeCheckpointer(MLFlowLogger):
 
 
 class GPUUsage(Callback):
-    """Track and log GPU usage."""
+    """Log GPU memory usage at logging step."""
 
     def __init__(self, log_freq=50):
         """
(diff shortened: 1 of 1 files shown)


meshnets/utils/callbacks.py
```diff
@@ -51,6 +51,32 @@ def finalize(self, status: str) -> None:
 class GPUUsage(Callback):
     """Track and log GPU usage."""
 
+    def __init__(self, log_freq=50):
```
IvanPombo, line 54: Add the class definition comment a better description and difference from this one to the one below.

meshnets/utils/callbacks.py
```diff
@@ -51,6 +51,32 @@ def finalize(self, status: str) -> None:
...
+        """
+        Args:
+            log_freq: The frequency, in batches, at which to log GPU usage.
+        """
+        self.log_freq = log_freq
+        self.gpu_usage_history = {}
+
+    def on_batch_end(self, trainer, pl_module):
+        if (trainer.global_step +
+                1) % self.log_freq == 0 or trainer.is_last_batch:
```
IvanPombo, line 64: Shall we log at the first batch as well?

meshnets/utils/callbacks.py
```diff
@@ -51,6 +51,32 @@ def finalize(self, status: str) -> None:
...
+            try:
+                all_gpus = GPUtil.getGPUs()
+                for gpu in all_gpus:
+                    trainer.logger.log_metrics(
```
IvanPombo, line 68: How will this look? All gpus together or separate info about them?

IvanPombo (approved):
LGTM!

meshnets/utils/callbacks.py
VictorGillioz: Each GPU is logged individually so there is a metric "GPU_NAME_memory_usage" for each of them.

meshnets/utils/callbacks.py
VictorGillioz: Screenshot 2023-08-11 at 11 20 52

I tried to follow lightning implementation in the class LoggerConnector that Trainer uses. This way the callback logs are aligned with the training step logs.

meshnets/utils/callbacks.py
VictorGillioz: Done, let me know if this is better

sarmento:
LGTM.
Does this also show up on MLFlow. Can you add a printscreen?

meshnets/utils/callbacks.py
IvanPombo: Seems good!

sarmento (approved):

AugustoPeres (approved):
LGTM

VictorGillioz:
Here goes the print screen @sarmento 

Screenshot 2023-08-14 at 11 50 05

AugustoPeres:
> Here goes the print screen @sarmento
> 
> Screenshot 2023-08-14 at 11 50 05

This is a very small job, right? Because the usage seems to be consistent at 0.07 usage. For the models you are training we should see more usage, no?

sarmento:
This is very cool!
Thank you so much.
A couple of things here that could potentially improve the readability here:

1) what is the unit? Should it be encoded in the name of the variable or can we add a tag/info string to the metric?

2) the name of the metric does not give any clue with respect to where is being taken from. Is this over all GPU's or over one specific GPU? If it is over one specific GPU, should the hostname also be encoded in the name of the variable (or in a tag)?

3) it is not clear what is the unit on the x axis?

Question: why do we get a zero and then a bump after n epochs/batches?

VictorGillioz:
@AugustoPeres : 

This is indeed a very small job to illustrate. I trained a small model on the smallest dataset we have for 10 epochs.

In almost all the training I did, the usage seems to stay constant between batches. Here it is 7% of the GPU memory.

VictorGillioz:
@sarmento : 

1. This is percentage here. I am not sure MLFlow offers a way to log explicitly what the metric is but I could add this information to the metric’s name if needed.
2. This is the total memory used for one GPU, with no detail about who or what process uses the memory. In this case, we track the memory usage for GPU-f699…
3. The unit selected for the x axis here is « Step ». This can be changed on the left panel and is taken care by MLFlow

sarmento:
OK! I was looking at MLFlow docs and it seems that there is nothing ele I can attach to a metric, so the name has to encode everything. That is a bit limiting...

@AugustoPeres : do you know if metrics can have meta-tags?

AugustoPeres:
> OK! I was looking at MLFlow docs and it seems that there is nothing ele I can attach to a metric, so the name has to encode everything. That is a bit limiting...
> 
> @AugustoPeres : do you know if metrics can have meta-tags?

We can have tags for the experiment not the metrics. To the best of my knowledge there is no way to attach tags to the [metric class](https://mlflow.org/docs/latest/_modules/mlflow/entities/metric.html#Metric)

AugustoPeres:
Actually, @VictorGillioz, and this is on me for not noticing, can you log the metrics with the name in [this format](https://github.com/inductiva/inductiva-ml-codebase/blob/f0a4b94823327a1a430fe9d82156af62ee725936/inductiva_ml/callbacks.py#L37). This is easier for me when retriving usage from mlflow

VictorGillioz:
@AugustoPeres you mean the metric name right?

sarmento:
> > @AugustoPeres you mean the metric name right?
> 
> yes

Let's just make sure we are not duplicating code, ok? 
Please try to integrate / re-use as much code as possible.
Thanks!

PR #34, opened 2023-08-14
new lightning wrapper params

VictorGillioz:
PR for a new dynamic in the lightning wrapper parameters. We now give the model Class and model parameters to the wrapper.

This has multiple advantages:
- Modularity between wrapper and model is preserved
- Wrapper and model are instantiated together
- The wrapper is able to save the model parameter as hyperparameters
- The instantiated model is not saved to the checkpoints (this usually throws a warning due to bad memory use)
- We can still load the wrapper without providing the hyperparameters

### DIFF
--- a/meshnets/utils/model_training.py
+++ b/meshnets/utils/model_training.py
@@ -78,19 +78,18 @@ def train_model(config, experiment_config, train_dataset, validation_dataset):
                                    persistent_workers=True,
                                    shuffle=False)
 
-    # Define the model
-    model = MeshGraphNet(
+    # Define the wrapper and the model
+    lightning_wrapper = MGNLightningWrapper(
+        MeshGraphNet,
         node_features_size=train_dataset.dataset.num_node_features,
         edge_features_size=train_dataset.dataset.num_edge_features,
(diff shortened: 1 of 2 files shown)


AugustoPeres (approved):
LGTM!! Thanks Victor, this should work nicely

meshnets/utils/model_training.py
```diff
@@ -78,28 +78,25 @@ def train_model(config, experiment_config, train_dataset, validation_dataset):
                                    persistent_workers=True,
                                    shuffle=False)
 
-    # Define the model
-    model = MeshGraphNet(
+    # Define the wrapper and the model
+    lightning_wrapper = MGNLightningWrapper(
```
sarmento, line 82: try just importing meshnets and expand the the path here to include all the packages.
This allow us to detect redundancies in the name of the pacakages and in the name of the classes.

sarmento (approved):
This looks good.
As you go along, try NOT to import classes directly.

VictorGillioz:
Thanks @sarmento for the suggestion. I saw @hpenedones ’s mail and I am planning to change all the imports in the project soon. I will do a separate PR for that

PR #37, opened 2023-08-18
added main file for model evaluation

VictorGillioz:
PR for evaluating a model on a validation dataset, can be either in or out of distribution.

### DIFF
--- a/meshnets/eval.py
+++ b/meshnets/eval.py
@@ -1,0 +1,65 @@
+"""The main file for model evaluation."""
+
+import os
+
+from absl import app
+from absl import flags
+import pytorch_lightning as pl
+import torch
+from torch.utils.data import random_split
+from torch_geometric.loader import DataLoader
+
+from meshnets.utils.datasets import FromDiskGeometricDataset
+from meshnets.utils import model_loading
+
+FLAGS = flags.FLAGS
+
+# Random seed flag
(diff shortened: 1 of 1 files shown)


meshnets/eval.py
```diff
@@ -0,0 +1,65 @@
...
+import torch
+from torch.utils.data import random_split
+from torch_geometric.loader import DataLoader
+
+from meshnets.utils.datasets import FromDiskGeometricDataset
+from meshnets.utils import model_loading
+
+FLAGS = flags.FLAGS
+
+# Random seed flag
```
AugustoPeres, line 17: Remove these comments. The string description of the Flag must be explanation enough

meshnets/eval.py
```diff
@@ -0,0 +1,62 @@
...
+    random_seed = FLAGS.random_seed
+    if random_seed is not None:
+        torch.manual_seed(random_seed)
+
+    dataset = FromDiskGeometricDataset(FLAGS.data_dir)
+    _, validation_dataset = random_split(
+        dataset, [FLAGS.train_split, FLAGS.validation_split])
+
+    validation_dataloader = DataLoader(validation_dataset,
+                                       batch_size=8,
```
fabiocruz, line 48: Should this be configurable? Or at least set in a global var?

fabiocruz (approved):

PR #38, opened 2023-08-18
added main file for pred visualization

VictorGillioz:
PR for visualising a model prediction on a random mesh from a validation dataset, can be in or out of distribution

plot

### DIFF
--- a/meshnets/utils/data_visualization.py
+++ b/meshnets/utils/data_visualization.py
@@ -32,8 +32,10 @@ def _validate_scalars(scalars: np.array) -> np.array:
         return scalars
     elif scalars.ndim == 2 and scalars.shape[1] == 1:
         return scalars.flatten()
-    elif scalars.ndim > 1:
-        raise ValueError("Argument 'scalars' must have a single dimension.")
+    else:
+        raise ValueError(
+            f"Argument 'scalars' has invalid dimensions : {scalars.shape}. "
+            'Dimensions must be (None,) or (None, 1).')
 
 
 def plot_mesh_with_scalars(mesh_path: str,
--- a/meshnets/utils/data_visualization.py
+++ b/meshnets/utils/data_visualization.py
@@ -2,7 +2,6 @@
 
 from typing import Union
 from typing import Tuple
-import warnings
 
 import numpy as np
 import pyvista as pv
@@ -24,6 +23,19 @@ def plot_mesh(mesh_path: str,
               screenshot=screenshot)
 
 
+def _validate_scalars(scalars: np.array) -> np.array:
+
+    if scalars is None:
+        raise TypeError("Argument 'scalars' is None.")
+
+    if scalars.ndim == 1:
+        return scalars
+    elif scalars.ndim == 2 and scalars.shape[1] == 1:
+        return scalars.flatten()
+    elif scalars.ndim > 1:
+        raise ValueError("Argument 'scalars' must have a single dimension.")
+
+
 def plot_mesh_with_scalars(mesh_path: str,
                            scalars: np.array,
                            clim: Union[Tuple[float, float], None] = None,
@@ -35,9 +47,7 @@ def plot_mesh_with_scalars(mesh_path: str,
     mesh = pv.read(mesh_path)
     mesh = mesh.rotate_z(rot_z)
 
-    if scalars is None:
-        warnings.warn(
-            "Argument 'scalars' is None. Displaying with active scalar array.")
+    scalars = _validate_scalars(scalars)
 
     mesh.plot(scalars=scalars,
               show_edges=True,
@@ -58,6 +68,9 @@ def plot_mesh_comparison(mesh_path: str,
 
     mesh = pv.read(mesh_path)
     mesh = mesh.rotate_z(rot_z)
+
+    ground_truth = _validate_scalars(ground_truth)
+    prediction = _validate_scalars(prediction)
 
     plotter = pv.Plotter(shape=(1, 3), off_screen=off_screen)
 
--- a/meshnets/visualize.py
+++ b/meshnets/visualize.py
@@ -18,7 +18,7 @@ FLAGS = flags.FLAGS
 flags.DEFINE_integer('random_seed', 21,
                      'The seed to initialize the random number generator.')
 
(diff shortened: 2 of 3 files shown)


meshnets/visualize.py
```diff
@@ -0,0 +1,87 @@
...
+import torch
+from torch.utils.data import random_split
+
+from meshnets.utils.datasets import FromDiskGeometricDataset
+from meshnets.utils import data_visualization
+from meshnets.utils import model_loading
+
+FLAGS = flags.FLAGS
+
+# Random seed flag
```
AugustoPeres, line 18: Remove these comments

meshnets/visualize.py
```diff
@@ -0,0 +1,87 @@
...
+    sample_idx = np.random.choice(validation_dataset.indices)
+
+    mesh_path = dataset.get_mesh_path(sample_idx)
+    graph_path = dataset.get_graph_path(sample_idx)
+
+    graph = torch.load(graph_path)
+    graph = wrapper.normalize_labels(graph)
+    normalized_groundtruth = graph.y.detach().numpy()
+
+    prediction = wrapper(graph).detach().numpy()
```
AugustoPeres, line 70: I think that you may want to call `wrapper.eval()` to tell pytorch_lightning that this is to eval the model. This good practice as it will shutdown dropouts, use inference batch norm, .... Even if such layer do not exist, this is good practice regardless

When using the trainer.eval(model, dataloader) it is not required as lightning will do it under the hood. But here I think it is.

AugustoPeres: Also, instead of `wrapper(graph).detach()` use `with torch.no_grad()`. This will prevent tensor being put in the computational graph, saving memory and eliminating the need of then removing them from there

AugustoPeres (approved):
LGTM, left just a few comments

meshnets/utils/data_visualization.py
```diff
@@ -24,6 +23,19 @@ def plot_mesh(mesh_path: str,
...
+
+    if scalars is None:
+        raise TypeError("Argument 'scalars' is None.")
+
+    if scalars.ndim == 1:
+        return scalars
+    elif scalars.ndim == 2 and scalars.shape[1] == 1:
+        return scalars.flatten()
+    elif scalars.ndim > 1:
+        raise ValueError("Argument 'scalars' must have a single dimension.")
```
fabiocruz, line 36: I think this should be an `else` clause (or even better, the clause should be omitted). The message itself is also not accurate, since by line 33 `scalars.ndim` can be 2. Rephrase the error message.

meshnets/visualize.py
```diff
@@ -0,0 +1,85 @@
...
+
+flags.DEFINE_string('tracking_uri', None,
+                    'The tracking URI for the MLFlow experiments.')
+
+flags.DEFINE_string('run_id', None, 'The run id of the experiment to load.')
+
+flags.DEFINE_integer('checkpoint', 0,
+                     'The checkpoint to load from the experiment.')
+
+flags.DEFINE_bool('start_xvfb', False, 'Wheter to start xvfb or not.')
```
fabiocruz, line 37: Typo, "whether"

meshnets/visualize.py
```diff
@@ -0,0 +1,85 @@
...
+
+flags.DEFINE_string('save_dir', 'img', 'Directory where to save the plots.')
+
+flags.DEFINE_string('file_name', 'plot.png', 'The name of the the plot file.')
```
fabiocruz, line 41: Why not have one flag only with `output_file_path`?

meshnets/visualize.py
```diff
@@ -0,0 +1,85 @@
...
+
+    data_visualization.plot_mesh_comparison(mesh_path,
+                                            normalized_groundtruth,
+                                            prediction,
+                                            clim=None,
+                                            rot_z=180,
+                                            off_screen=False,
+                                            screenshot=os.path.join(
+                                                FLAGS.save_dir,
+                                                FLAGS.file_name))
```
fabiocruz, line 81: No reason for these two flags to exist. One would be enough.

meshnets/visualize.py
```diff
@@ -0,0 +1,85 @@
...
+                                                   FLAGS.checkpoint)
+
+    sample_idx = np.random.choice(validation_dataset.indices)
+
+    mesh_path = dataset.get_mesh_path(sample_idx)
+    graph_path = dataset.get_graph_path(sample_idx)
+
+    with torch.no_grad():
+        graph = torch.load(graph_path)
+        graph = wrapper.normalize_labels(graph)
```
fabiocruz, line 65: Is `normalize_labels` a staticmethod of the wrapper class? It should be stored elsewhere, if so.

meshnets/visualize.py
```diff
@@ -0,0 +1,85 @@
...
+        normalized_groundtruth = graph.y
+
```
fabiocruz, line 67: Can you provide more context to what's happening in each of these lines (in comments to the PR, to start with).

fabiocruz (approved):
LGTM. I have minor(ish) comments only.

meshnets/visualize.py
VictorGillioz: It is not a static method, as the normalisation depends on the label statistics the wrapper has been initialised with.

meshnets/visualize.py
VictorGillioz: We want to compare the groundtruth, obtained from the raw pressure field and normalized according to dataset statistics, and the prediction, infered from the graph representation by the model trained with normalized labels.

meshnets/visualize.py
fabiocruz: I see, thanls.

meshnets/visualize.py
fabiocruz: Thanks.

meshnets/utils/data_visualization.py
VictorGillioz: You're right, I changed it to an else clause and changed the error message. Why do you think the clause should be omitted? I wrote this method because pyvista allows faulty plots when the provided array does not conform to their required dimensions. This leads to plots that looks plausible but do not represent the correct data.

PR #39, opened 2023-08-22
changed geometric data passing in model

VictorGillioz:
This PR changes the argument passing throughout the whole MGN model.

Except for the main model, the forward methods do not receive `Batch` anymore, but the detailed arguments s.a. `x` (the node features), `edge_attr` (the edge features), `edge_index` (the list of edge indices). 

This seems to be closer to standard practice, and avoids accidents resulting from `Data`/`Batch` being unmutable.

Additional implication is: replacement of `Sequential` by `ModuleList` in the processor, because sequential can only receive single arguments.

Small unrelated addition: added `__repr__` method in the Processor layer class because the default `MessagePassing` method does not detail inner modules.

### DIFF
--- a/meshnets/modules/decoder.py
+++ b/meshnets/modules/decoder.py
@@ -1,7 +1,6 @@
 """"Define the GraphDecoder class."""
 
 import torch
-from torch_geometric.data import Batch
 
 from meshnets.modules.mlp import MLP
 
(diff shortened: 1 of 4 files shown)


AugustoPeres:
> Additional implication is: replacement of Sequential by ModuleList in the processor, because sequential can only receive single arguments as input.

There is a geometric sequential that can receive more than single arguments. Look [here](https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html) to see if it is something that will be useful for you. But I know you can do something like:

```python
model = Sequential('x, edge_index, edge_features',  # this are the outputs of the model
[
    (MyLayer, 'x, edge_index -> x'), # this is a tuple with the layer, the inputs, and the output signature
])
```

This is not a request, it is just something for you to be aware in case you want to keep using `Sequential`

AugustoPeres (approved):
LGTM!! I think that this is better, thanks Victor

PR #42, opened 2023-08-28
Multi val

VictorGillioz:
PR to support multiple validation datasets during training.

This is done by loading multiple datasets in the main file, and giving them as a list to the main training method together with a list of the dataset names for improved logging.

newplot
newplot (1)
newplot (2)


sarmento:
@VictorGillioz: This is great!

Let me make a few more suggestions just to increase readability, analysing results. All these things are mostly related with generating the plots, not so much with general infra-structure / data piping

1) Give option to generate a zoom view over just a range of the epochs. Obviously the first 100 or 200 epochs have a very high error, and this compresses the detail when you are showing what happens at epoch 1500-2000. In other words, it would be great to see what happens at the beginning and at the end of the training separately.

2) Related to this, maybe you want to zoom in on a certain region of the y axis. If the values go above a certain max y, then you simply don't show them. Again, this helps zooming in on a certain part of the training curves.

3) also related with this: Maybe we could "solve" the two previous points by showing a logarithm scale on the y axis.

4) Caption box: it would be great to always show training loss first, and then you show the list of the validation losses. Also, it my help if you use a different type of line for training and validation losses (e.g. full vs ticks or points).

5) The labels of the datasets are not very informative. I understand that you use those labels for logging, but would it be possible to map them to human readable names when you plot. Also, related to that: what is the resolution of the training data? Maybe you also want to point that out in the corresponding label.

6) Titles are important. They really help understanding the purpose of the plot. Eg. "Losses over validation sets of different resolution". Something like this really helps the reader understand your point. 

7) Also, you need to add units to the lost. Is this RMS computed over all point in the grid?

sarmento:
Final note: it seems from the plots that validation losses are still going down *very fast*.
I am a bit intrigued by that. Have you left it training for 10k epochs? How long are these training procedures taking?

Should we train with an even bigger (and hopefully more diverse) dataset, so that the model learns more in each epoch? 

Have we tested  the models in CADillac dataset? (how many points do objects in that dataset have?)

VictorGillioz:
Thanks @sarmento for the extensive feedback! To answer your questions:

1. This is already possible with MLFlow
2. This too :)
3. Also possible
4. The metrics order is changeable, but I don't think the type of line is
5. The metric label is currently "val_loss_dataset_name/dataloader_idx_X": the first part is in my opinion quite readable and depends on the dataset name which includes mesh resolution in our case (e.g 2k), the suffix is added by lightning and cannot be removed in the version we use
6. The plots are exported from MLFlow and I don't think there is an option to add a title by default, but a title can of course be added when using the image
7. I don't think MLFlow allows to indicate units, the loss displayed here is MSE over all points, I can add that is the loss name if you want

VictorGillioz:
Regarding the loss, I have trained for more epochs in other cases, this run was made to illustrate this PR only.

Bigger datasets might be a good option to improve results, this would need to be tested.

The models have not been tested on CADillac yet

sarmento:
Ah! This is taken directly from MLFLow!!! OK! I thought you were getting data from MLFlow and then calling matplotlib or something similar to generate the plots. That would give you a lot more control, but I think this is fine for now.

AugustoPeres:
> @VictorGillioz: This is great!

> 5. The labels of the datasets are not very informative. I understand that you use those labels for logging, but would it be possible to map them to human readable names when you plot. Also, related to that: what is the resolution of the training data? Maybe you also want to point that out in the corresponding label.
> 6. Titles are important. They really help understanding the purpose of the plot. Eg. "Losses over validation sets of different resolution". Something like this really helps the reader understand your point.
> 7. Also, you need to add units to the lost. Is this RMS computed over all point in the grid?

I am not sure that we want to add all of this features to mlflow. The only way to add titles, new labels and different line types to these visualizations is to make the plot ourselves and then logging that plot to mlflow which will make comparing different runs more difficult. As opposed to doing:

```python
mlflow.log_metric('metric_name', value, step)
```

However, we can still plot everything that @sarmento suggested but outside mlflow. Mlflow stores as csv all the metrics that were logged there. Therefore, we can have a python script that makes plots with all these features and run it for the runs for which we want to create some type of report, blog post, or any other publication

sarmento:
> > @VictorGillioz: This is great!
> 
> > 5. The labels of the datasets are not very informative. I understand that you use those labels for logging, but would it be possible to map them to human readable names when you plot. Also, related to that: what is the resolution of the training data? Maybe you also want to point that out in the corresponding label.
> > 6. Titles are important. They really help understanding the purpose of the plot. Eg. "Losses over validation sets of different resolution". Something like this really helps the reader understand your point.
> > 7. Also, you need to add units to the lost. Is this RMS computed over all point in the grid?
> 
> I am not sure that we want to add all of this features to mlflow. The only way to add titles, new labels and different line types to these visualizations is to make the plot ourselves and then logging that plot to mlflow which will make comparing different runs more difficult. As opposed to doing:
> 
> ```python
> mlflow.log_metric('metric_name', value, step)
> ```
> 
> However, we can still plot everything that @sarmento suggested but outside mlflow. Mlflow stores as csv all the metrics that were logged there. Therefore, we can have a python script that makes plots with all these features and run it for the runs for which we want to create some type of report, blog post, or any other publication

I feel that this would be the best alternative.

meshnets/tune.py
```diff
@@ -72,10 +74,23 @@ def main(_):
...
         dataset, [FLAGS.train_split, FLAGS.validation_split])
 
+    val_datasets_names = [dataset_name]
+    val_datasets = [validation_dataset]
+
+    for val_data_dir in FLAGS.val_data_dirs:
+        dataset_name = os.path.basename(val_data_dir)
+        dataset = FromDiskGeometricDataset(val_data_dir)
+        _, validation_dataset = random_split(
+            dataset, [FLAGS.train_split, FLAGS.validation_split])
```
AugustoPeres, line 89: Why do we split the data. I was already expecting that the directories in `val_data_dirs` to contain exactly the data that we wanted to use for validation.

AugustoPeres (approved):
LGTM

meshnets/tune.py
VictorGillioz: `val_data_dir` gives the path to the datasets we want to use for validation, we still split them to avoid evaluating on the whole dataset

Issue #44, opened 2023-08-29
Make meshnets packageble 


AugustoPeres:
At the moment there is no way to install meshnets as package so that it can be used outside the main `/meshnets` folder. Ideally we would have this type of folder structure:

```
.
├── # This is the front page of the repo meshnets
├── setup.py
└── meshnets/
    ├── __init__.py
    ├── modules/
    │   ├── decoder.py
    │   └── __init__.py
    ├── datatypes
    └── utils
```

Where the setup file will be something like:

```python
"""Setup script"""
from setuptools import setup, find_packages

setup(name='meshenets', version='1.0', packages=find_packages())
```

And the `__init__.py` would be something like:

```python
# meshenets/modules/__init__.py
from .decoder import *
```

This would allow users to install meshenets by:

1. Cloning the repository;
2. `cd` into the repository;
3. Running `pip install -r requirements.txt`;
4. Running `pip install -e .`

Now meshnets can even be used outside of the repo by simply using:

```python
# some/other/part/of/my/computer/example.py
import meshnets
```

Addressing this issue will consist on creating the `setup.py` file and the `__init__.py_files`. This would even allow to move the python scripts such as [`meshnets/train.py`](https://github.com/inductiva/meshnets/blob/main/meshnets/train.py) to a folder `scripts/train.py`

IvanPombo:
This seems good! Should we also discuss the other ml package where some of the code here is coming from? Or are you thinking of this being a completely independent one?

AugustoPeres:
> This seems good! Should we also discuss the other ml package where some of the code here is coming from? Or are you thinking of this being a completely independent one?

The other ML packages are already implemented like this. I think that mesh nets should be separate.

sarmento:
Thank you for this. I am really happy that we are moving in the direction of having high quality open-source contributions. @hpenedones : does it make sense to have a blog post about this?

hpenedones:
> Thank you for this. I am really happy that we are moving in the direction of having high quality open-source contributions. @hpenedones : does it make sense to have a blog post about this?

Yes, I think it does make sense to have a blog post. Victor will be less available in the coming weeks, but we can work on it among the rest of us.

PR #45, opened 2023-08-29
Making meshnets a package.

AugustoPeres:
Addresses issue #44.

Previews:

Install
image

Import and python autocomplete
image

### DIFF
--- a/meshnets/modules/__init__.py
+++ b/meshnets/modules/__init__.py
@@ -1,0 +1,7 @@
+"""Init"""
+from .decoder import *
+from .encoder import *
+from .lightning_wrapper import *
+from .mlp import *
+from .model import *
+from .processor import *
--- a/meshnets/modules/__init__.py
+++ b/meshnets/modules/__init__.py
@@ -1,7 +1,7 @@
 """Init"""
-from .decoder import *
-from .encoder import *
-from .lightning_wrapper import *
-from .mlp import *
-from .model import *
-from .processor import *
+from .decoder import GraphDecoder
+from .encoder import GraphEncoder
+from .lightning_wrapper import MGNLightningWrapper
+from .mlp import MLP
+from .model import MeshGraphNet
+from .processor import MGNProcessorLayer, MGNProcessor
--- a/meshnets/utils/__init__.py
+++ b/meshnets/utils/__init__.py
@@ -1,8 +1,13 @@
 """Init"""
-from .callbacks import *
-from .data_loading import *
-from .data_processing import *
-from .data_visualization import *
-from .datasets import *
-from .model_loading import *
-from .model_training import *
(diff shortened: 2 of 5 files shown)


meshnets/modules/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""Init"""
+from .decoder import *
```
IvanPombo, line 2: Should we be more explicit?

IvanPombo (approved):
LGTM! Just left two minor comments.

meshnets/modules/__init__.py
AugustoPeres: Only if there is something that we specifically do not want to be importable (I think)

fabiocruz: This isn't a good practice, since it omits what is being imported. Import all submodules you want instead.

fabiocruz (changes requested):
Wildcard imports, `from <module> import *` should be avoided for clarity. See e.g. [this](https://peps.python.org/pep-0008/#imports).

AugustoPeres:
> Wildcard imports, `from <module> import *` should be avoided for clarity. See e.g. [this](https://peps.python.org/pep-0008/#imports).

I was under the impression that this did not apply to `__init__` files. But I was wrong. Will change it now

meshnets/utils/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+                        GPUUsageMean, GradientNorm, GeometricBatchSize)
+from .data_loading import (load_edge_mesh_pv, edges_from_meshio_mesh,
+                           load_edge_mesh_meshio, load_triangle_mesh)
+from .data_processing import (edge_mesh_to_graph, triangle_mesh_to_graph,
+                              mesh_file_to_graph_data)
+from .data_visualization import (plot_mesh, plot_mesh_with_scalars,
+                                 plot_mesh_comparison)
+from .datasets import FromDiskGeometricDataset
+from .model_loading import load_model_from_mlflow
+from .model_training import train_model
```
fabiocruz, line 13: I think in this particular file you want to do e.g. `from . import callbacks`, as not to expose all utilities equally, e.g. `meshnets.utils.FromDiskGeometricDataset` and `meshnets.utils.GradientNorm`. I think it's useful to keep the submodule names, e.g. `meshnets.utils.datasets.FromDiskGeometricDataset` for context.

fabiocruz (approved):
Looking good, although I think you should import things slightly different in the `utils` submodule. See my comment below.

meshnets/utils/__init__.py
AugustoPeres: Done
image

AugustoPeres:
Going to merge: @VictorGillioz please take a look at this and issue #45 when you can

PR #46, opened 2023-09-15
Update readme

AugustoPeres:
This pull request updates the readme file. It was added:

* A brief abstract for the project. This was actually written by @IvanPombo and copy pasted by me;
* A brief intro to mlflow and how to use it to manage experiments;
* Instructions on where to get the data;
* Instructions on how to train the models

All the instruction in the readme where tested on a brand new virtual environment and everything should work if the order of commands is followed.

**Edited**: If you want to look at the rendered markdown go [here](https://github.com/inductiva/meshnets/tree/update_readme)


README.md
```diff
@@ -1,172 +1,124 @@
...
-  However, if you want to run `yapf` and `pylint` locally, simply install them
-  via `pip`:
+If you do not have your own data you can download one of our datasets
+[here](https://drive.google.com/u/0/uc?id=1T5wXz-igXjrztLt6xgb4CZx4tGwXK8DX&export=download). This
+particular dataset has 1000 examples of objects:
 
-  ```bash
-  pip install yapf pylint
-  ```
+![alt_image](readme_images/dataset_example.png)
```
IvanPombo, line 42: Let's improve these images afterwards...

README.md
```diff
@@ -1,172 +1,124 @@
...
 
-  ```bash
-  pip install pytest
-  ```
+The next step is to pre-process the dataset. To do this just run the
+script:
 
-  and run:
+```bash
+pytohn process --data_dir=<Path to your dataset>
```
IvanPombo, line 52: typo *python

IvanPombo: This can be different at times... In my case it will be python3

README.md
```diff
@@ -1,172 +1,124 @@
...
 ```
 
-and run
+The very next step is to create a virtual environment. This will solve
+any clashes with the library versions used here and anything else that
+might be installed in your own system:
 
 ```bash
-sphinx-build -b html docs docs/_build
+python3 -m venv .env
```
IvanPombo, line 87: See this for a comment above.

README.md
```diff
@@ -1,172 +1,124 @@
...
+```
 
-  ```bash
-  pytest
-  ```
+You will see `.pt` objects appearing next to the `.vtk` objects. These
+are the torch graphs already constructed.
 
-  in your base directory.
+## Mlflow
```
IvanPombo, line 58: If this is an extra step for those really want to track everything, I think this could below the installation, since it won't be required for it.

README.md
```diff
@@ -1,172 +1,124 @@
...
+[mlflow](https://mlflow.org/docs/latest/python_api/mlflow.html) to
+keep track of our experiments. We have our own remote server to which
+we can log everything from any computer which was created following
+the instructions in [this
+tutorial](https://towardsdatascience.com/managing-your-machine-learning-experiments-with-mlflow-1cd6ee21996e).
 
-## Documentation
+This is a nice feature to have but it is not required in any way to
+run our code. Everything will be logged locally by default. That is,
+all experiments will be logged to the folder `mlruns` created in the
```
IvanPombo, line 69: Maybe change the default name of this folder. Is it possible?

README.md
```diff
@@ -1,172 +1,124 @@
...
+all the requirements of the project using:
 
 ```bash
-pip install sphinx-autobuild
+pip install --upgrade pip
+pip install -r requirements.txt
 ```
 
-and run in your base directory
+Next, because `meshents` is actually packaged we can install it using:
```
IvanPombo, line 99: typo meshnets

README.md
```diff
@@ -1,172 +1,124 @@
...
-To automate all the setup commands above, we have created a Conda
-environment config file that you can use for your convenience.
-
-Make sure the first line of the conda.yml file contains a name related to 
-your project, such as:
-
-```
-name: startaproject_env
+```python
+python train.py --data_dir=<Path to your dataset>
```
IvanPombo, line 111: If I follow all of these steps will this actually work? Notice, that we haven't said anywhere to split the datasets.

README.md
AugustoPeres: > This can be different at times... In my case it will be python3

Even when you create the virtual env? I need to create the virtual env with `python3` but after that it  is just `python`

README.md
AugustoPeres: This is the mlflow default, it is not specified by us. The user would need to create an environment variable to change this.

README.md
AugustoPeres: image
It is split in the main script.

IvanPombo: That the path to the validation dataset is the name of the folder that then stays with the validation samples, right?

README.md
IvanPombo: I mean that it depends of the people using. But leave it like this and next I will test this approach and check :)

README.md
```diff
@@ -1,172 +1,124 @@
-[![Python package](https://github.com//actions/workflows/python-package.yml/badge.svg)](https://github.com//actions/workflows/python-package.yml)
+# Meshnets project
 
-[![Documentation Status](https://readthedocs.com/projects/inductiva-research-labs-/badge/?version=latest&token=a7a7344c5bc9e511ca715449f94612e4bb7c2dff4c90931778b9c85dbbf5480c)](https://inductiva-research-labs-.readthedocs-hosted.com/en/latest/?badge=latest)
+While virtual wind tunnels are a much cheaper and flexible option than
```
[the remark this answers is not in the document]
AugustoPeres: > Is the code limited to training data from windtunnel simulations, or it works for any mesh to mesh problem, with scalar values in each mesh triangle?

I am starting to look at the code in depth so I do not want to miss lead you but if I had to guess:

1. The models and training logic work for any mesh/graph;
2. The data pre-processing is specific to wind tunnel problems.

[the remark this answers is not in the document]
VictorGillioz: Hi, I would like to confirm and complete Augusto's comment.

1. The model and training procedure work on any kind of graph data; the training procedure can be slightly modified to also work with other models
2. The data pre-processing is specific to our wind tunnel problems, but could be slightly modified to suit other mesh pre-processing in a similar MGN fashion

PR #47, opened 2023-09-25
Better dataset splitting

AugustoPeres:
This pull request removes an unnecessary and and bug prone flag (FLAGS.validation_split) in the visualization script.
Now we just need to give the dataset the split ratio.

**Note**: I think it may be even better to remove this all-together and just give the script a path to validation data which is, by default split on disc.

### DIFF
--- a/meshnets/eval.py
+++ b/meshnets/eval.py
@@ -1,5 +1,4 @@
 """The main file for model evaluation."""
-
 import os
 
 from absl import app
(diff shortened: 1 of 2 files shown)


IvanPombo (approved):
LGTM! What do you mean by your suggestion to split on disk? Before doing this altogether?

AugustoPeres:
> LGTM! What do you mean by your suggestion to split on disk? Before doing this altogether?

I think we should just have the data organized in a folder like:

```
10k_simulations/
      train/
      validation/
      test/
```

No need to carry around a random seed to make sure the split is the same when doing inference

PR #48, opened 2023-09-26
Visualizing relative error

AugustoPeres:
Adds the option to visualize the relative error in the visualization script.

We then call this function twice. The last time we clip the colorbar to the range [0, 1] to visualize where the error is greater than 1.

### DIFF
--- a/meshnets/utils/data_visualization.py
+++ b/meshnets/utils/data_visualization.py
@@ -107,5 +107,35 @@ def plot_mesh_comparison(mesh_path: str,
                      cmap='RdBu_r',
                      clim=clim,
                      copy_mesh=True)
+    plotter.show(screenshot=screenshot)
+
+
+def plot_relative_error(mesh_path: str,
+                        ground_truth: np.array,
+                        prediction: np.array,
+                        clim: Union[Tuple[float, float], None] = None,
+                        rot_z: int = 0,
+                        off_screen: bool = False,
+                        screenshot: Union[str, None] = None):
+    mesh = pv.read(mesh_path)
+    mesh = mesh.rotate_z(rot_z)
+
+    ground_truth = _validate_scalars(ground_truth)
+    prediction = _validate_scalars(prediction)
+
+    plotter = pv.Plotter(shape=(1, 1), off_screen=off_screen)
+
+    # Relative error
+    difference = prediction - ground_truth
+    relative_error = np.abs(difference / (ground_truth + 1e-7))
+
+    plotter.subplot(0, 0)
+    plotter.add_text('Relative Error')
+    plotter.add_mesh(mesh,
+                     scalars=relative_error,
+                     show_edges=True,
+                     roughness=1,
+                     clim=clim,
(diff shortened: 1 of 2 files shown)


meshnets/utils/data_visualization.py
```diff
@@ -107,5 +107,35 @@ def plot_mesh_comparison(mesh_path: str,
...
+    plotter.show(screenshot=screenshot)
+
+
+def plot_relative_error(mesh_path: str,
+                        ground_truth: np.array,
+                        prediction: np.array,
+                        clim: Union[Tuple[float, float], None] = None,
+                        rot_z: int = 0,
+                        off_screen: bool = False,
+                        screenshot: Union[str, None] = None):
```
IvanPombo, line 119: Could you add here a description of what you meant in the PR to change the color bar.

meshnets/utils/data_visualization.py
```diff
@@ -107,5 +107,35 @@ def plot_mesh_comparison(mesh_path: str,
...
                      clim=clim,
                      copy_mesh=True)
+    plotter.show(screenshot=screenshot)
+
+
+def plot_relative_error(mesh_path: str,
+                        ground_truth: np.array,
+                        prediction: np.array,
+                        clim: Union[Tuple[float, float], None] = None,
+                        rot_z: int = 0,
```
IvanPombo, line 117: *rotate_z

meshnets/utils/data_visualization.py
```diff
@@ -107,5 +107,35 @@ def plot_mesh_comparison(mesh_path: str,
...
+                        off_screen: bool = False,
+                        screenshot: Union[str, None] = None):
+    mesh = pv.read(mesh_path)
+    mesh = mesh.rotate_z(rot_z)
+
+    ground_truth = _validate_scalars(ground_truth)
+    prediction = _validate_scalars(prediction)
+
+    plotter = pv.Plotter(shape=(1, 1), off_screen=off_screen)
```
IvanPombo, line 126: Why you need the shape and subplot at all?

IvanPombo (approved):
LGTM, after you explain a little bit more what you mean by plotting a second time...

meshnets/utils/data_visualization.py
AugustoPeres: Basically there are outliers in the predictions that have very large errors. Therefore, we we use the colorbar with range [0, max_error] it becomes very hard to differentiate between the smaller errors as they will be assigned indistinguishable colors. Thus, I make a second plot with the colorbar values clipped to the range [0, 1] so that we can take a look at the smaller errors

meshnets/utils/data_visualization.py
AugustoPeres: You are right, I don't

meshnets/utils/data_visualization.py
IvanPombo: But I thought that this will always be less 1, since it is the relative error.

PR #49, opened 2023-09-27
Api data generation

AugustoPeres:
This pull request adds scripts to generate data from .obj files to the repository. A more detailed description of the pull request and how to use the scripts can be found in the readme committed here as well.

### DIFF
--- a/data_generation/README.md
+++ b/data_generation/README.md
@@ -1,0 +1,34 @@
+# Generate data
+
+Use this scripts to generate data with the Inductiva API.
+
+The first thing that you need to do is export your api key as an
+environment variable:
+
+```bash
+export INDUCTIVA_API_KEY = <Your API Key>
+```
+
+This will prevent accidentally committing it to github since now we
+only need to do:
+
+```
+inductiva.api_key = os.environ['INDUCTIVA_API_KEY']
+```
+
+
+To submit the tasks use the script `submit_tasks.py`. This will submit
+the tasks to the API and save a `.txt` file with their IDs.
+
+To then fetch the output of these tasks you can run the script
+`download_tasks.py`. If no `--download_dir` is given the script will
+only monitor the tasks and output something like:
+
+```bash
+I0927 10:09:20.328027 140566996619328 download_tasks.py:26] Number of tasks: 30
+I0927 10:09:23.044083 140566996619328 download_tasks.py:47] Tasks still running: 0
+I0927 10:09:23.044179 140566996619328 download_tasks.py:48] Tasks successfully completed: 30
+I0927 10:09:23.044209 140566996619328 download_tasks.py:50] Tasks failed: 0
+I0927 10:09:23.044236 140566996619328 download_tasks.py:51] Tasks killed: 0
+I0927 10:09:23.044259 140566996619328 download_tasks.py:52] Other tasks: 0
+```
--- a/data_generation/README.md
+++ b/data_generation/README.md
@@ -18,7 +18,10 @@ inductiva.api_key = os.environ['INDUCTIVA_API_KEY']
 
 
 To submit the tasks use the script `submit_tasks.py`. This will submit
-the tasks to the API and save a `.txt` file with their IDs.
+the tasks to the API and save a `.txt` file with their IDs. To run the
+script you will `.vtk` objects over which to run the simulations. Once
+you have the `.vtk` objects over which to run the simulation run
+`python submit_tasks.py --input_dataset=<path to .vtk objects>`.
 
 To then fetch the output of these tasks you can run the script
 `download_tasks.py`. If no `--download_dir` is given the script will
--- a/data_generation/download_tasks.py
+++ b/data_generation/download_tasks.py
@@ -1,0 +1,66 @@
+"""This file downloads the requested tasks"""
+import os
+
+from absl import app
+from absl import flags
+from absl import logging
+
+import inductiva
+
+FLAGS = flags.FLAGS
+
+flags.DEFINE_string('path_to_sim_ids', None,
+                    'The path to the simulation ids txt files.')
+
+flags.DEFINE_string(
+    'download_dir', None,
+    'Where to download the data. If None, the data is not downloaded.')
+
+
+def main(_):
+    inductiva.api_key = os.environ['INDUCTIVA_API_KEY']
+
+    # Read the simulation ids
+    with open(FLAGS.path_to_sim_ids, 'r') as f:
+        task_ids = f.read().splitlines()
+    logging.info('Number of tasks: %s', len(task_ids))
+
+    tasks_still_running = []
+    tasks_successfully_completed = []
+    tasks_failed = []
+    tasks_killed = []
+    other_tasks = []
+    for task_id in task_ids:
+        task = inductiva.tasks.Task(task_id)
+        status = task.get_status()
+        if status == 'success':
+            tasks_successfully_completed.append(task_id)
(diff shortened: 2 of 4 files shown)


data_generation/README.md
```diff
@@ -0,0 +1,37 @@
+# Generate data
+
+Use this scripts to generate data with the Inductiva API.
```
IvanPombo, line 3: *these

data_generation/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+```bash
+export INDUCTIVA_API_KEY = <Your API Key>
+```
+
+This will prevent accidentally committing it to github since now we
+only need to do:
+
+```
+inductiva.api_key = os.environ['INDUCTIVA_API_KEY']
```
IvanPombo, line 16: Technically, this won't be necessary, since when doing import inductiva this will be done automatically.

data_generation/README.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+
+To submit the tasks use the script `submit_tasks.py`. This will submit
+the tasks to the API and save a `.txt` file with their IDs. To run the
+script you will `.vtk` objects over which to run the simulations. Once
```
IvanPombo, line 22: you will... something missing here.

IvanPombo: I am confused on this part... Do you mean the `.obj` files?

data_generation/README.md
```diff
@@ -0,0 +1,37 @@
...
+you have the `.vtk` objects over which to run the simulation run
+`python submit_tasks.py --input_dataset=<path to .vtk objects>`.
+
+To then fetch the output of these tasks you can run the script
+`download_tasks.py`. If no `--download_dir` is given the script will
```
IvanPombo, line 27: If the script is named download_tasks.py then it should download the data to a default bucket, no?

You could have a script to monitor things, if you desire.

data_generation/download_tasks.py
```diff
@@ -0,0 +1,66 @@
...
+flags.DEFINE_string("path_to_sim_ids", None,
+                    "The path to the simulation ids txt files.")
+
+flags.DEFINE_string(
+    "download_dir", None,
+    "Where to download the data. If None, the data is not downloaded.")
+
+
+def main(_):
+    inductiva.api_key = os.environ["INDUCTIVA_API_KEY"]
```
IvanPombo, line 21: Remove.

data_generation/download_tasks.py
```diff
@@ -0,0 +1,66 @@
...
+    with open(FLAGS.path_to_sim_ids, "r", encoding="utf-8") as f:
+        task_ids = f.read().splitlines()
+    logging.info("Number of tasks: %s", len(task_ids))
+
+    tasks_still_running = []
+    tasks_successfully_completed = []
+    tasks_failed = []
+    tasks_killed = []
+    other_tasks = []
+    for task_id in task_ids:
```
IvanPombo, line 33: I know you want this for two purposes, but I think it would be better to split it into a monitor scripting and a download_script. Or just change the name of the script to mention reflect the possibilities.

data_generation/download_tasks.py
```diff
@@ -0,0 +1,66 @@
...
+flags.DEFINE_string(
+    "download_dir", None,
+    "Where to download the data. If None, the data is not downloaded.")
+
+
+def main(_):
+    inductiva.api_key = os.environ["INDUCTIVA_API_KEY"]
+
+    # Read the simulation ids
+    with open(FLAGS.path_to_sim_ids, "r", encoding="utf-8") as f:
```
IvanPombo, line 24: I feel that this should be a feature we should have... Given a set of task_ids, fetch all task objects in one go. But given the simplicity I am not sure... Wdyt as a user?

data_generation/download_tasks.py
```diff
@@ -0,0 +1,66 @@
...
+    logging.info("Tasks still running: %s", len(tasks_still_running))
+    logging.info("Tasks successfully completed: %s",
+                 len(tasks_successfully_completed))
+    logging.info("Tasks failed: %s", len(tasks_failed))
+    logging.info("Tasks killed: %s", len(tasks_killed))
+    logging.info("Other tasks: %s", len(other_tasks))
+
+    if FLAGS.download_dir is not None:
+        for task_id in tasks_successfully_completed:
+            task = inductiva.tasks.Task(task_id)
```
IvanPombo, line 56: Why not save the task object here, instead of the task_id, since you will actually do the above all the time?

data_generation/submit_tasks.py
```diff
@@ -0,0 +1,80 @@
...
+        for object_path in object_paths
+    ]
+
+    # Save all the tasks ids to a txt file.
+    with open(os.path.join(FLAGS.output_dataset, "task_ids.txt"),
+              "w",
+              encoding="utf-8") as f:
+        for task_id in task_ids:
+            f.write(str(task_id.id) + "\n")
+
```
IvanPombo, line 77: My only concern here, is that you are not terminating the machines :) Could we add this on the "monitor.py" for example?

IvanPombo (changes requested):
Left a couple of comments to start iterating. The main thing I would like to understand and receive your feedback on, is the way you managing and downloading the tasks.

data_generation/README.md
AugustoPeres: Is this feature available already?

IvanPombo: I believe so! But you may need to set it manually on `.bashrc` or respective file for your terminal of use.

data_generation/README.md
AugustoPeres: Yeah, I though about that, but then we need three scripts on our workflow. Where here we can just use 2 and then, when ready, download the data

data_generation/submit_tasks.py
AugustoPeres: Yup, good point. My only question is how to terminate the machine groups that correspond to the tasks. For example, suppose that I ran the `submit_tasks.py` script twice to generate two datasets. How do I know which machine group is running which tasks so that I do not accidentally terminate the wrong one?

We could, just like for the tasks, store the ID of the machine group. Is there a way to do this?

data_generation/README.md
AugustoPeres: Will split in 3 scripts

data_generation/download_tasks.py
AugustoPeres: I feel it is simple enough to fetch them one by one

data_generation/submit_tasks.py
IvanPombo: I think you could create a json file with the machine group name and the task_ids. You have the attribute `machine_group.name` and then you can fetch the machine to kill it with the `inductiva.resources.get`... I think finding with the name is not possible yet - check here https://github.com/inductiva/inductiva/tree/main/inductiva/resources. But if this is a feature you think is valuable, I advise you to open an issue with it, regarding the way you want to use it - then we move forward from there.

Leave this as is now, but remember to kill the machines :)

data_generation/README.md
IvanPombo: I believe the way to monitor things will be much more used, than the download outputs. Since simulations may take a long time, it is valuable to call it several times.

data_generation/download_tasks.py
IvanPombo: Alright!

IvanPombo (approved):
LGTM! I just left a suggestion to improve the text in the README.md.

PR #52, opened 2023-10-02
Removing deprecated code

AugustoPeres:
This pull request removes code that is no longer in use through out the repository. In the screen shots we can see the results of search across the repository for the functions/classes that were removed

## Data types

The datatypes are no longer in use and we can see no references to them in the code other than in their respective `__init__.py` files:

image
image

## Data loading

There were also function in `data_loading.py` that are no longer in use. All these functions were related to the library `meshio`.

image
image
image

## Testing

After removing the functions I ran the script `process.py` which handles data pre-processing and everything seemed in working order


IvanPombo (approved):
I believe this were from the beginning of the project and they were unused... Thanks @AugustoPeres!

PR #55, opened 2023-10-03
Filter out folders that do not a .pt file

AugustoPeres:
## Motivation

Sometimes our datasets contain simulation folders that do not have a torch graph. This is usually due to the tasks corresponding to that simulation having failed. The way the dataset was implemented before raised an error when trying to access those non-existing files:

```python
    def get(self, idx: int) -> Data:
        """Get an element from the dataset at the given index.
        
        The only things that exist in RAM at this point are the paths 
        to the samples. This method loads and returns the sample from disk."""

        graph_path = self.get_graph_path(idx)
        return torch.load(graph_path)
```

But, the way this was implemented, it could be the case that `get_graph_path` was returning a non-existing path

```python
    def get_graph_path(self, idx: int) -> str:
        """Get the path to the graph file at the given index."""

        sample_path = self.get_sample_path(idx)
        graph_path = os.path.join(sample_path, self.graph_file_name)

        if not os.path.isfile(graph_path):
            warnings.warn(f"File '{graph_path}' does not exist.")

        return graph_path
```

This pull request filters and warns the user immediately in the `__init__` method if there are folders with no graphs.

**Note**: This is not a replacement for making sure that our datasets have all the intended simulations. But, while we work on that front, this was added to prevent training scripts from crashing. Moreover, since a warning is being raised it is fine that we filter empty folders since it will always be apparent to the user that their data has problems

## Testing

To test this I ran it on a test dataset with just 10 examples, one of which consists of an empty  folder:

image

Iterating over the dataset yields the 9 non-empty examples:

image

### DIFF
--- a/meshnets/utils/datasets.py
+++ b/meshnets/utils/datasets.py
@@ -48,14 +48,24 @@ class FromDiskGeometricDataset(Dataset):
         super().__init__(*args, **kwargs)
 
         self.root_dir = data_dir
+
+        self.mesh_file_name = mesh_file_name
+        self.graph_file_name = graph_file_name
+
         samples = [
             sample for sample in os.listdir(data_dir)
             if os.path.isdir(os.path.join(self.root_dir, sample))
         ]
-        self.samples = sorted(samples)
 
-        self.mesh_file_name = mesh_file_name
-        self.graph_file_name = graph_file_name
+        # Filter out samples that do not have a `.pt` file.
+        for sample in samples:
+            graph_path = os.path.join(self.root_dir, sample,
+                                      self.graph_file_name)
+            if not os.path.isfile(graph_path):
+                warnings.warn(f"Sample '{sample}' does not have a graph file.")
+                samples.remove(sample)
+
+        self.samples = sorted(samples)
 
     def len(self) -> int:
         """Return the number of samples in the dataset."""
--- a/meshnets/utils/datasets.py
+++ b/meshnets/utils/datasets.py
@@ -92,9 +92,6 @@ class FromDiskGeometricDataset(Dataset):
         sample_path = self.get_sample_path(idx)
         graph_path = os.path.join(sample_path, self.graph_file_name)
 
-        if not os.path.isfile(graph_path):
-            warnings.warn(f"File '{graph_path}' does not exist.")
-
         return graph_path
 
     def get(self, idx: int) -> Data:


IvanPombo (approved):
This looks good to merge! However, I don't really understand why in general this should be the behaviour... Shouldn't we first filter the simulation data, before passing here? It should be a separate thing? 

The idea is that we will generate a large sample of simulations and a small percentage of them can fail, for whatever reason, therefore we should immediately drop this folders there... Maybe not even download... I think the inherent problem here is that we are assuming a certain ordered for the simulations, which for me doesn't make much sense, since the data is random... Could explain me this last thign?

AugustoPeres:
> This looks good to merge! However, I don't really understand why in general this should be the behaviour... Shouldn't we first filter the simulation data, before passing here? It should be a separate thing?
> 
> The idea is that we will generate a large sample of simulations and a small percentage of them can fail, for whatever reason, therefore we should immediately drop this folders there... Maybe not even download... I think the inherent problem here is that we are assuming a certain ordered for the simulations, which for me doesn't make much sense, since the data is random... Could explain me this last thign?

I have folders with the structure:

```
├── 1696000210446528819
│   ├── object.obj
├── 1696000211442312370
│   ├── object.obj
│   ├── pressure_field_interpolated.vtk
│   ├── pressure_field.pt
│   └── pressure_field.vtk
├── 1696000212643164059
│   ├── object.obj
│   ├── pressure_field_interpolated.vtk
│   ├── pressure_field.pt
│   └── pressure_field.vtk
├── 1696000213656639981
│   ├── object.obj
│   ├── pressure_field_interpolated.vtk
│   ├── pressure_field.pt
│   └── pressure_field.vtk
└── sim_info.json
```

Notice that, for the first folder, the simulation failed so we do not have a `.vtk` file. I could filter that folder in two ways:

1. Copy the successful simulations to another folder;
2. Delete the failed simulation.

If we follow 1. we now have two sources of truth for the same dataset. Additionally we need more disk space. Finally, if someone else wanted to run the code they would either have to pick the filtered dataset or make sure that they manually filter out the failed simulations.

If we opt for 2 we now loose information about the `.obj` objects for which the simulation failed. This will make it harder to try and run the simulations again to complete this missing data. By keeping the objects we could now create a `resubmit_tasks.py`, path it the file `sim_info.json`, and run those simulations again to complete the dataset.

I agree that this should not replace proper data preprocessing and I am working on filling in the gaps in the dataset but, for now, this allows me to train models with the data as is. Moreover, this small check in the `__init__` adds no undesired behaviors (we just filter out incomplete data anyways) and, since warning are raised, in the future it is unlikely that we may be filtering out things without realizing.

IvanPombo:
> > This looks good to merge! However, I don't really understand why in general this should be the behaviour... Shouldn't we first filter the simulation data, before passing here? It should be a separate thing?
> > The idea is that we will generate a large sample of simulations and a small percentage of them can fail, for whatever reason, therefore we should immediately drop this folders there... Maybe not even download... I think the inherent problem here is that we are assuming a certain ordered for the simulations, which for me doesn't make much sense, since the data is random... Could explain me this last thign?
> 
> I have folders with the structure:
> 
> ```
> ├── 1696000210446528819
> │   ├── object.obj
> ├── 1696000211442312370
> │   ├── object.obj
> │   ├── pressure_field_interpolated.vtk
> │   ├── pressure_field.pt
> │   └── pressure_field.vtk
> ├── 1696000212643164059
> │   ├── object.obj
> │   ├── pressure_field_interpolated.vtk
> │   ├── pressure_field.pt
> │   └── pressure_field.vtk
> ├── 1696000213656639981
> │   ├── object.obj
> │   ├── pressure_field_interpolated.vtk
> │   ├── pressure_field.pt
> │   └── pressure_field.vtk
> └── sim_info.json
> ```
> 
> Notice that, for the first folder, the simulation failed so we do not have a `.vtk` file. I could filter that folder in two ways:
> 
> 1. Copy the successful simulations to another folder;
> 2. Delete the failed simulation.
> 
> If we follow 1. we now have two sources of truth for the same dataset. Additionally we need more disk space. Finally, if someone else wanted to run the code they would either have to pick the filtered dataset or make sure that they manually filter out the failed simulations.
> 
> If we opt for 2 we now loose information about the `.obj` objects for which the simulation failed. This will make it harder to try and run the simulations again to complete this missing data. By keeping the objects we could now create a `resubmit_tasks.py`, path it the file `sim_info.json`, and run those simulations again to complete the dataset.
> 
> I agree that this should not replace proper data preprocessing and I am working on filling in the gaps in the dataset but, for now, this allows me to train models with the data as is. Moreover, this small check in the `__init__` adds no undesired behaviors (we just filter out incomplete data anyways) and, since warning are raised, in the future it is unlikely that we may be filtering out things without realizing.

Maybe I am being naive in terms of the DL pipeline, but you could just have the successful simulations folder in your dataset and pass the failed to another folder. Single source of truth and no duplication. When running new simulations with the failed ones, you could add to the successful dataset.

However, indeed this allows you to start training and maybe later we can discuss the appropriate methodology to combine the simulation pipeline with the training procedure! Thanks @AugustoPeres!

sarmento:
Hey everyone!
I am not sure if I am getting all the thread right, but think this is a great opportunity to improve the API UX -- which should always be a priority. 

So, if I understand the basics, the source of the problem is, either:

1) that some simulations didn't finish -- for whatever reason -- and that passed undetected until now, so we donwloaded the partial results that are not good for the subsequent ML step

2) the simulation was ok byt downloading itself failed (e.g. connection dropped or whatever).

I don't think nr 2 is very likely. Nr 1 probably what happened.

Since generating data at scale is one of the most important use cases of the API I think we need to discuss a bit how we deal with nr 1. As a user, it should be a very simple way for knowing if the simulation was ok, and if simulations don't finish well, there should be a way of checking that without having to check each of the files. Are we missing any feature in the API?

@hpenedones wdyt?

sarmento:
> > > This looks good to merge! However, I don't really understand why in general this should be the behaviour... Shouldn't we first filter the simulation data, before passing here? It should be a separate thing?
> > > The idea is that we will generate a large sample of simulations and a small percentage of them can fail, for whatever reason, therefore we should immediately drop this folders there... Maybe not even download... I think the inherent problem here is that we are assuming a certain ordered for the simulations, which for me doesn't make much sense, since the data is random... Could explain me this last thign?
> > 
> > 
> > I have folders with the structure:
> > ```
> > ├── 1696000210446528819
> > │   ├── object.obj
> > ├── 1696000211442312370
> > │   ├── object.obj
> > │   ├── pressure_field_interpolated.vtk
> > │   ├── pressure_field.pt
> > │   └── pressure_field.vtk
> > ├── 1696000212643164059
> > │   ├── object.obj
> > │   ├── pressure_field_interpolated.vtk
> > │   ├── pressure_field.pt
> > │   └── pressure_field.vtk
> > ├── 1696000213656639981
> > │   ├── object.obj
> > │   ├── pressure_field_interpolated.vtk
> > │   ├── pressure_field.pt
> > │   └── pressure_field.vtk
> > └── sim_info.json
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Notice that, for the first folder, the simulation failed so we do not have a `.vtk` file. I could filter that folder in two ways:
> > 
> > 1. Copy the successful simulations to another folder;
> > 2. Delete the failed simulation.
> > 
> > If we follow 1. we now have two sources of truth for the same dataset. Additionally we need more disk space. Finally, if someone else wanted to run the code they would either have to pick the filtered dataset or make sure that they manually filter out the failed simulations.
> > If we opt for 2 we now loose information about the `.obj` objects for which the simulation failed. This will make it harder to try and run the simulations again to complete this missing data. By keeping the objects we could now create a `resubmit_tasks.py`, path it the file `sim_info.json`, and run those simulations again to complete the dataset.
> > I agree that this should not replace proper data preprocessing and I am working on filling in the gaps in the dataset but, for now, this allows me to train models with the data as is. Moreover, this small check in the `__init__` adds no undesired behaviors (we just filter out incomplete data anyways) and, since warning are raised, in the future it is unlikely that we may be filtering out things without realizing.
> 
> Maybe I am being naive in terms of the DL pipeline, but you could just have the successful simulations folder in your dataset and pass the failed to another folder. Single source of truth and no duplication. When running new simulations with the failed ones, you could add to the successful dataset.
> 
> However, indeed this allows you to start training and maybe later we can discuss the appropriate methodology to combine the simulation pipeline with the training procedure! Thanks @AugustoPeres!

Please don't drop this ball. It's very important for us to reflect on this and see what needs to be done in the API to create the best possible experience. At this point, it's more important to get the API UX right than it is to train a specific ML model.

AugustoPeres:
> Hey everyone! I am not sure if I am getting all the thread right, but think this is a great opportunity to improve the API UX -- which should always be a priority.
> 
> So, if I understand the basics, the source of the problem is, either:
> 
> 1. that some simulations didn't finish -- for whatever reason -- and that passed undetected until now, so we donwloaded the partial results that are not good for the subsequent ML step
> 2. the simulation was ok byt downloading itself failed (e.g. connection dropped or whatever).
> 
> I don't think nr 2 is very likely. Nr 1 probably what happened.
> 
> Since generating data at scale is one of the most important use cases of the API I think we need to discuss a bit how we deal with nr 1. As a user, it should be a very simple way for knowing if the simulation was ok, and if simulations don't finish well, there should be a way of checking that without having to check each of the files. Are we missing any feature in the API?
> 
> @hpenedones wdyt?

You are right, number 1 happend.

> As a user, it should be a very simple way for knowing if the simulation was ok

It is, If you take a look at the script [`monitor_tasks.py`](https://github.com/inductiva/meshnets/blob/4e138bf79ffe13b0e7fa4eb23b67e1826009a330/data_generation/monitor_tasks.py#L33) we can see that we can look at the task status just from the ID. 

To miss in the gaps, an ideal feature would be something like:

```python
task.resubmit()
```

To try and run the tasks that failed again

AugustoPeres:
> > > This looks good to merge! However, I don't really understand why in general this should be the behaviour... Shouldn't we first filter the simulation data, before passing here? It should be a separate thing?
> > > The idea is that we will generate a large sample of simulations and a small percentage of them can fail, for whatever reason, therefore we should immediately drop this folders there... Maybe not even download... I think the inherent problem here is that we are assuming a certain ordered for the simulations, which for me doesn't make much sense, since the data is random... Could explain me this last thign?
> > 
> > 
> > I have folders with the structure:
> > ```
> > ├── 1696000210446528819
> > │   ├── object.obj
> > ├── 1696000211442312370
> > │   ├── object.obj
> > │   ├── pressure_field_interpolated.vtk
> > │   ├── pressure_field.pt
> > │   └── pressure_field.vtk
> > ├── 1696000212643164059
> > │   ├── object.obj
> > │   ├── pressure_field_interpolated.vtk
> > │   ├── pressure_field.pt
> > │   └── pressure_field.vtk
> > ├── 1696000213656639981
> > │   ├── object.obj
> > │   ├── pressure_field_interpolated.vtk
> > │   ├── pressure_field.pt
> > │   └── pressure_field.vtk
> > └── sim_info.json
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Notice that, for the first folder, the simulation failed so we do not have a `.vtk` file. I could filter that folder in two ways:
> > 
> > 1. Copy the successful simulations to another folder;
> > 2. Delete the failed simulation.
> > 
> > If we follow 1. we now have two sources of truth for the same dataset. Additionally we need more disk space. Finally, if someone else wanted to run the code they would either have to pick the filtered dataset or make sure that they manually filter out the failed simulations.
> > If we opt for 2 we now loose information about the `.obj` objects for which the simulation failed. This will make it harder to try and run the simulations again to complete this missing data. By keeping the objects we could now create a `resubmit_tasks.py`, path it the file `sim_info.json`, and run those simulations again to complete the dataset.
> > I agree that this should not replace proper data preprocessing and I am working on filling in the gaps in the dataset but, for now, this allows me to train models with the data as is. Moreover, this small check in the `__init__` adds no undesired behaviors (we just filter out incomplete data anyways) and, since warning are raised, in the future it is unlikely that we may be filtering out things without realizing.
> 
> Maybe I am being naive in terms of the DL pipeline, but you could just have the successful simulations folder in your dataset and pass the failed to another folder. Single source of truth and no duplication. When running new simulations with the failed ones, you could add to the successful dataset.
> 
> However, indeed this allows you to start training and maybe later we can discuss the appropriate methodology to combine the simulation pipeline with the training procedure! Thanks @AugustoPeres!

Thanks Ivan. Will merge and later we can discuss how to better handle the data pipeline through the API

PR #56, opened 2023-10-04
Changing default mesh_file_name name

AugustoPeres:
Now we mainly use the interpolates .vtk files so I am changing the defaults

### DIFF
--- a/meshnets/utils/datasets.py
+++ b/meshnets/utils/datasets.py
@@ -42,8 +42,8 @@ class FromDiskGeometricDataset(Dataset):
     def __init__(self,
                  data_dir: str,
                  *args,
-                 mesh_file_name: str = 'pressure_field.vtk',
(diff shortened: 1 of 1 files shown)


meshnets/utils/datasets.py
```diff
@@ -42,7 +42,7 @@ class FromDiskGeometricDataset(Dataset):
     def __init__(self,
                  data_dir: str,
                  *args,
-                 mesh_file_name: str = 'pressure_field.vtk',
+                 mesh_file_name: str = 'pressure_field_interpolated.vtk',
```
IvanPombo, line 45: I guess the graph we will call it the same?

IvanPombo (approved):

meshnets/utils/datasets.py
AugustoPeres: Yes, since we only have one graph for training. But I see your point. We should maybe also rename the graphs in the pre-processing step

PR #58, opened 2023-10-12
Added the option to terminate machine groups on system crash and on C^c

AugustoPeres:
## Description
This pull request adds error handling and termination of the machine group in case of script crashes or exceptions. It registers a signal handler for Ctrl+C interruptions and uses a try-except block to catch any unhandled exceptions.

@IvanPombo is there a way we could have this functionality by default integrated on the API?

## Changes Made
- Added a signal handler for Ctrl+C interruptions to terminate the machine group gracefully.
- Wrapped the main logic of the script in a try-except block to catch any unhandled exceptions.
- Terminated the machine group in the except block

## Testing

A tested the Ctrl+C functionality by manually aborting the script while monitoring the machines groups with:

```python
inductiva.resources.machine_groups.list()
```

I introduced a division by zero (now removed) in the scripts to test catching exceptions while monitoring the launched machine groups.

### DIFF
--- a/data_generation/submit_tasks.py
+++ b/data_generation/submit_tasks.py
@@ -1,8 +1,11 @@
 """This file uses the inductiva API to sumbit tasks"""
 from absl import app
 from absl import flags
+from absl import logging
 
 import os
+import signal
+import sys
 import shutil
 import json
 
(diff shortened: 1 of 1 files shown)


sarmento:
@AugustoPeres : this is a great functionality, and this helps not having resources on if the script fails. I think this is a good feature. 

@hpenedones @luispcunha @begaiym-k how can we make this part of the machine group either as default behaviour or as an optional behaviour that the user turns on (or off). Ideally, this is the type of things that the client should do for you automatically, not requiring the user write this signal handling code themselves.

It's a great feature in terms of simplicity and making sure costs don't explode by accident.

PR #61, opened 2023-10-17
Changes to processing

AugustoPeres:
## Motivation

This pull request implements changes to the data processing to make it more compatible with hugging face datasets.
Before we were processing the data directly to torch graphs. While, after these changes we can now process the data into four different numpy arrays, edge_index, edge_atributes, node_features and targets.

This will make loading data through HuggingFace much easier. We can simply create a HuggingFace dataset with columns: 'x', 'edge_index', 'edge_attr', 'y' and then create a `collate_fn` function to create the dataloader. Something like this:

```python
def collate_fn(examples):
    graphs = [torch_geometric.data.Data(x=torch.tensor(example) for example in examples]
    return torch_geometric.data.Batch.from_data_list(graphs)

data_loader = torch.utils.data.DataLoader(meshnets_dataset_train, collate_fn=collate_fn, batch_size=2)
```

## Changes

To do this I create three new function. One to create the edges and the edge attributes, another to create the node features, another to create the target predictions. I tested these functions on a single triangle mesh created with the script:

```python
import pyvista as pv
import numpy as np

# Add the cube mesh to the plotter
cells = [3, 0, 1, 2]
points = [[1, 0, 0], [0, 1, 0], [0, 0, 1]]

mesh = pv.PolyData(points, cells)

# To the mesh add a vector 'p' with 3 components
mesh['p'] = np.random.rand(mesh.n_points, 1)

# Create a PyVista plotter
plotter = pv.Plotter()

plotter.add_mesh(mesh.points, color='r')
plotter.add_mesh(mesh.extract_all_edges(), color='k')
# Show the plotter window
plotter.show()

mesh.save('triangle.vtk')
```

Which yields this:

image

And then loaded it using the new functions:

image

## Additional changes

I also changed the process.py script to incorporate these changes and removed a deprecated function that was not used throughout the repo.

## On deprecated code

There are now a few functions that will be deprecated soon but which I have left at the moment for compatibility motives with my current workflow.

### DIFF
--- a/meshnets/process.py
+++ b/meshnets/process.py
@@ -10,7 +10,7 @@ import torch
 import pyvista as pv
 import numpy as np
 
-from meshnets.utils import data_processing
+from meshnets.utils import data_processing, data_loading
 
 FLAGS = flags.FLAGS
 
@@ -26,6 +26,14 @@ flags.DEFINE_string('original_mesh_name', 'object.obj',
                     'Name of the original mesh.')
 flags.DEFINE_string('wind_vector_name', 'flow_velocity.json',
                     'The name of the file containing the wind vector.')
+flags.DEFINE_string('edge_index_name', 'edge_index.npy',
+                    'Name of the edge index file.')
+flags.DEFINE_string('edge_features_name', 'edge_features.npy',
+                    'Name of the edge attribute file.')
+flags.DEFINE_string('node_features_name', 'node_features.npy',
+                    'Name of the node attribute file.')
+flags.DEFINE_string('wind_pressures_name', 'wind_pressures.npy',
+                    'Name of the wind pressures file.')
 flags.DEFINE_string('torch_graph_name', 'pressure_field.pt',
                     'Name of the torch graph.')
 flags.DEFINE_list(
@@ -75,8 +83,11 @@ def main(_):
             folder, FLAGS.interpolated_pressure_field_name)
         interpolated_mesh.save(interpolated_mesh_path)
 
-        edge_index, edge_attr = data_processing.make_edge_index_and_features(
-            interpolated_mesh_path)
+        nodes, edges, pressures = data_loading.load_edge_mesh_pv(
+            interpolated_mesh_path, load_pressure=True)
+
+        edge_index, edge_features =\
+            data_processing.make_edge_index_and_features(nodes, edges)
 
         wind_vector_path = os.path.join(folder, FLAGS.wind_vector_name)
         if os.path.exists(wind_vector_path):
@@ -84,16 +95,15 @@ def main(_):
                 wind_vector = json.load(f)
         else:
             wind_vector = FLAGS.default_wind_vector
-        node_features = data_processing.make_node_features(
-            interpolated_mesh_path, wind_vector)
+        node_features = data_processing.make_node_features(nodes, wind_vector)
 
-        target = data_processing.make_target(interpolated_mesh_path)
+        wind_pressures = np.array(pressures, dtype=np.float32)
 
         # Save as numpy arrays.
-        np.save(os.path.join(folder, 'edge_index.npy'), edge_index)
-        np.save(os.path.join(folder, 'edge_attr.npy'), edge_attr)
-        np.save(os.path.join(folder, 'node_features.npy'), node_features)
-        np.save(os.path.join(folder, 'target.npy'), target)
+        np.save(os.path.join(folder, FLAGS.edge_index_name), edge_index)
+        np.save(os.path.join(folder, FLAGS.edge_features_name), edge_features)
+        np.save(os.path.join(folder, FLAGS.node_features_name), node_features)
+        np.save(os.path.join(folder, FLAGS.wind_pressures_name), wind_pressures)
 
         # TODO(augusto): Remove this in the future. Leaving it here for
         # compatibility with the old code.
--- a/meshnets/process.py
+++ b/meshnets/process.py
@@ -1,5 +1,6 @@
 """The main file for data processing."""
 import os
+import json
 import warnings
 
 from absl import app
@@ -7,6 +8,7 @@ from absl import flags
 
 import torch
 import pyvista as pv
+import numpy as np
 
 from meshnets.utils import data_processing
 
@@ -22,16 +24,15 @@ flags.DEFINE_string('interpolated_pressure_field_name',
                     'Name of the interpolated pressure field.')
 flags.DEFINE_string('original_mesh_name', 'object.obj',
                     'Name of the original mesh.')
+flags.DEFINE_string('wind_vector_name', 'flow_velocity.json',
+                    'The name of the file containing the wind vector.')
 flags.DEFINE_string('torch_graph_name', 'pressure_field.pt',
                     'Name of the torch graph.')
+flags.DEFINE_list(
+    'default_wind_vector', [30., 0, 0],
+    'Default wind vector if none is present in the simulation folder.')
 
 flags.DEFINE_float('tolerance', 1.5, 'Tolerance for the mesh sampling.')
-
-# TODO: This will be deprecated soon as we will start training with
-# varying wind speeds. At the moment this is here just for support
-# while not metadata is added to the dataset containing the wind
-# speed.
-WIND_VECTOR = [30., 0, 0]
 
 
 def main(_):
@@ -53,13 +54,12 @@ def main(_):
     ]
 
     # Remove folders that do not contain a .vtk file.
-    for folder in sim_folders:
+    for folder in sim_folders.copy():
         if not os.path.exists(
                 os.path.join(folder, FLAGS.original_pressure_field_name)):
             warnings.warn(f'{folder} does not contain a .vtk file.')
             sim_folders.remove(folder)
 
-    # Convert the .vtk files to graph data.
     for folder in sim_folders:
         openfoam_mesh_path = os.path.join(folder,
                                           FLAGS.original_pressure_field_name)
@@ -75,9 +75,30 @@ def main(_):
             folder, FLAGS.interpolated_pressure_field_name)
         interpolated_mesh.save(interpolated_mesh_path)
 
-        graph_data = data_processing.mesh_file_to_graph_data(
-            interpolated_mesh_path, WIND_VECTOR, load_pressure=True)
+        edge_index, edge_attr = data_processing.make_edge_index_and_features(
+            interpolated_mesh_path)
 
+        wind_vector_path = os.path.join(folder, FLAGS.wind_vector_name)
+        if os.path.exists(wind_vector_path):
+            with open(wind_vector_path, encoding='utf-8') as f:
+                wind_vector = json.load(f)
+        else:
+            wind_vector = FLAGS.default_wind_vector
+        node_features = data_processing.make_node_features(
+            interpolated_mesh_path, wind_vector)
+
+        target = data_processing.make_target(interpolated_mesh_path)
+
+        # Save as numpy arrays.
+        np.save(os.path.join(folder, 'edge_index.npy'), edge_index)
+        np.save(os.path.join(folder, 'edge_attr.npy'), edge_attr)
+        np.save(os.path.join(folder, 'node_features.npy'), node_features)
+        np.save(os.path.join(folder, 'target.npy'), target)
+
+        # TODO(augusto): Remove this in the future. Leaving it here for
+        # compatibility with the old code.
+        graph_data = data_processing.mesh_file_to_graph_data(
+            interpolated_mesh_path, wind_vector, load_pressure=True)
         torch.save(graph_data, os.path.join(folder, FLAGS.torch_graph_name))
 
 
--- a/meshnets/utils/data_processing.py
+++ b/meshnets/utils/data_processing.py
@@ -8,7 +8,6 @@ import numpy as np
 import torch
 from torch_geometric.data import Data
 from torch_geometric.utils import to_undirected
-from torch_geometric.transforms.face_to_edge import FaceToEdge
 
 from meshnets.utils import data_loading
 
@@ -71,39 +70,6 @@ def edge_mesh_to_graph(node_coordinates: np.ndarray,
     return graph
 
 
-def triangle_mesh_to_graph(node_coordinates: np.ndarray,
-                           node_features: np.ndarray,
-                           faces: np.ndarray,
-                           node_labels: Optional[np.ndarray] = None) -> Data:
-    """Produce an undirected graph object from triangle mesh data and compute
-    its edge features.
-
-    Args shape:
-        node_coordinates: (nb_nodes, nb_space_dim)
-        node_features: (nb_nodes, nb_node_features)
-        faces: (nb_faces, 3)
-        node_labels: (nb_nodes, nb_node_labes)
-    
-    The graph includes node features, edges indices, edge attributes
-    and optionally node labels."""
-
-    node_coordinates = torch.tensor(node_coordinates, dtype=torch.float32)
-    node_features = torch.tensor(node_features, dtype=torch.float32)
-    faces = torch.tensor(faces.T, dtype=torch.int64)
-
-    if node_labels is not None:
-        node_labels = torch.tensor(node_labels, dtype=torch.float32)
-
-    mesh = Data(x=node_features,
-                face=faces,
-                y=node_labels,
-                pos=node_coordinates)
-    graph = FaceToEdge(remove_faces=True)(mesh)
-    graph = _compute_edge_attributes(graph, remove_pos=True)
-
-    return graph
-
-
 def mesh_file_to_graph_data(file_path: str,
                             wind_vector: Tuple[float],
                             load_pressure: bool = True,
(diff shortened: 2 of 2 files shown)


meshnets/utils/data_processing.py
```diff
@@ -138,3 +104,75 @@ def mesh_file_to_graph_data(file_path: str,
...
+    a wind vector. Return the edge index and the edge features.
+
+    Args:
+        file_path: Path to the mesh file
+        verbose: Whether to print information about the mesh file
+    Returns:
+        edge_index: Edge index for the mesh
+        edge_attr: Edge features for the mesh
+    """
+    nodes, edges, _ = data_loading.load_edge_mesh_pv(file_path,
```
hpenedones, line 146: seems like each of these 3 functions is loading the data from the file, and then discarding some of its info. Can we load it just once (avoid code redundancy AND make it faster) and then pass around the relevant data to this functions?

meshnets/process.py
```diff
@@ -75,9 +75,30 @@ def main(_):
...
+        node_features = data_processing.make_node_features(
+            interpolated_mesh_path, wind_vector)
 
+        target = data_processing.make_target(interpolated_mesh_path)
+
+        # Save as numpy arrays.
+        np.save(os.path.join(folder, 'edge_index.npy'), edge_index)
+        np.save(os.path.join(folder, 'edge_attr.npy'), edge_attr)
+        np.save(os.path.join(folder, 'node_features.npy'), node_features)
+        np.save(os.path.join(folder, 'target.npy'), target)
```
hpenedones, line 96: I think I wouldn't use such generic names, and probably it would not even hardcode in the name that one of them is the prediction target (unless that's a requirement?), because I can see the dataset being used in different ways. For example, when you add here the drag coefficient, or stream lines, perhaps those will be the targets rather than the pressure.
So, I would use names that are more descriptive of the contents of the data: such as pressure....

meshnets/process.py
```diff
@@ -75,9 +75,30 @@ def main(_):
...
+            wind_vector = FLAGS.default_wind_vector
+        node_features = data_processing.make_node_features(
+            interpolated_mesh_path, wind_vector)
 
+        target = data_processing.make_target(interpolated_mesh_path)
+
+        # Save as numpy arrays.
+        np.save(os.path.join(folder, 'edge_index.npy'), edge_index)
+        np.save(os.path.join(folder, 'edge_attr.npy'), edge_attr)
+        np.save(os.path.join(folder, 'node_features.npy'), node_features)
```
hpenedones, line 95: For node_features, I think we might be coupling what is a dataset generation step with what is an ML modelling choice. We want to store in the dataset what was the wind speed, ideally in a compact way, without making the assumption that this will be fed to the model via passing the same vector to every node in the graph. There are other choices of inputing that info the model that are different, so the dataset itself could try to be agnostic to those later choices.

Btw: is there a reason that for edges we use "attributes" but for nodes we use "features".

hpenedones (changes requested):
Hi Augusto, thanks, I left some initial comments. My view might be a little off (let me know) because I don't have full awareness of where these processing functions are plugging in the workflow? Is it before generating the HuggingFace dataset?

sarmento:
Following @hpenedones comments, is there a way of measuring the I/O performance related with these. I am afraid that all these layers may lead to hidden I/O overheads.


Unrelated to this: do we already have some info on how the current implementation of GNNs deals with the sparsity of these graphs? I think this was the biggest bottleneck for scaling, right?

meshnets/utils/data_processing.py
AugustoPeres: This a good point. Will change it on my next commit

meshnets/process.py
AugustoPeres: I will address the hard coded names. Regarding the names themselves I agree that target should be renamed to something like `wind_pressure`. The others seem to me to be pretty standard

meshnets/process.py
AugustoPeres: > Btw: is there a reason that for edges we use "attributes" but for nodes we use "features".

Not really. In pytorch they use `edge_attr` for the edge attributes/features that is why I used the "attributes" for the edges. However, for node "attributes/features" they use only `x`. To prevent such a generic name I just wrote "features" because that is what is most common in literature

> For node_features, I think we might be coupling what is a dataset generation step with what is an ML modelling choice. We want to store in the dataset what was the wind speed, ideally in a compact way, without making the assumption that this will be fed to the model via passing the same vector to every node in the graph.

I don't necessarily agree here. The scripts in `data_generation` are the ones responsible by data generation. This script handles the post processing of said data. I don't think that there is anything wrong with having a `process` script that processes the data in a way that will be useful in our pipeline later on even if it makes assumption on how to solve the problem. Here we just process the raw data and package it in a way that will be easily handled by HuggingFaces datasets. Most hugging face datasets have their data already packaged in some way that makes loading them easy. Take a look for example at [this](https://huggingface.co/datasets/graphs-datasets/MUTAG) dgraph dataset.

sarmento:
> > Following @hpenedones comments, is there a way of measuring the I/O performance related with these. I am afraid that all these layers may lead to hidden I/O overheads.
> 
> This is a processing script decoupled from training. Any I/O overhead that happens here will happen only once after processing the data.
> 
> > Unrelated to this: do we already have some info on how the current implementation of GNNs deals with the sparsity of these graphs? I think this was the biggest bottleneck for scaling, right?
> 
> The biggest bottleneck for training is the size of the computational graph used for autograd by pytorch. Which is a consequence of the large graphs in the dataset but not their sparsity


Can you explain better this:

> The biggest bottleneck for training is the size of the computational graph used for autograd by pytorch. Which is a consequence of the large graphs in the dataset but not their sparsity

Why would the size of the data graph have impact on the size of the computational graph. Shouldn't the computational graph be determined (mostly) by the number of layers, convolutions, etc?

sarmento:
> > Following @hpenedones comments, is there a way of measuring the I/O performance related with these. I am afraid that all these layers may lead to hidden I/O overheads.
> 
> This is a processing script decoupled from training. Any I/O overhead that happens here will happen only once after processing the data.

I am sorry: is this para of a pre-processing script?
This operation is not being (indirectly) called while training? The training process does not invoke in any way this operation, right?

AugustoPeres:
> > > Following @hpenedones comments, is there a way of measuring the I/O performance related with these. I am afraid that all these layers may lead to hidden I/O overheads.
> > 
> > 
> > This is a processing script decoupled from training. Any I/O overhead that happens here will happen only once after processing the data.
> 
> I am sorry: is this para of a pre-processing script? This operation is not being (indirectly) called while training? The training process does not invoke in any way this operation, right?

You are correct. The training does not invoke any of this operations

AugustoPeres:
> > > Following @hpenedones comments, is there a way of measuring the I/O performance related with these. I am afraid that all these layers may lead to hidden I/O overheads.
> > 
> > 
> > This is a processing script decoupled from training. Any I/O overhead that happens here will happen only once after processing the data.
> > > Unrelated to this: do we already have some info on how the current implementation of GNNs deals with the sparsity of these graphs? I think this was the biggest bottleneck for scaling, right?
> > 
> > 
> > The biggest bottleneck for training is the size of the computational graph used for autograd by pytorch. Which is a consequence of the large graphs in the dataset but not their sparsity
> 
> Can you explain better this:
> 
> > The biggest bottleneck for training is the size of the computational graph used for autograd by pytorch. Which is a consequence of the large graphs in the dataset but not their sparsity
> 
> Why would the size of the data graph have impact on the size of the computational graph. Shouldn't the computational graph be determined (mostly) by the number of layers, convolutions, etc?

It is affected by the number of layers as well. But also by the size of the inputs. Each node that is added to a graph corresponds to more message passing steps, hence larger computational graphs

sarmento:
> In fact, one thing we could do to avoid confusion is to have a separate repo, where we have code responsible for generating datasets using our API and packaging them under the HuggingFace interface. And the meshnets repo is focused on training mesh networks but it tries to be a little more agnostic/decoupled from data generation. Wdyt?

I like that!

> And the meshnets repo is focused on training mesh networks but it tries to be a little more agnostic/decoupled from data generation. Wdyt?

Yes. The meshnet code should be usable by other projects, and specific data dependencies should be separate.

sarmento:
> > > > Following @hpenedones comments, is there a way of measuring the I/O performance related with these. I am afraid that all these layers may lead to hidden I/O overheads.
> > > 
> > > 
> > > This is a processing script decoupled from training. Any I/O overhead that happens here will happen only once after processing the data.
> > > > Unrelated to this: do we already have some info on how the current implementation of GNNs deals with the sparsity of these graphs? I think this was the biggest bottleneck for scaling, right?
> > > 
> > > 
> > > The biggest bottleneck for training is the size of the computational graph used for autograd by pytorch. Which is a consequence of the large graphs in the dataset but not their sparsity
> > 
> > 
> > Can you explain better this:
> > > The biggest bottleneck for training is the size of the computational graph used for autograd by pytorch. Which is a consequence of the large graphs in the dataset but not their sparsity
> > 
> > 
> > Why would the size of the data graph have impact on the size of the computational graph. Shouldn't the computational graph be determined (mostly) by the number of layers, convolutions, etc?
> 
> It is affected by the number of layers as well. But also by the size of the inputs. Each node that is added to a graph corresponds to more message passing steps, hence larger computational graphs

I am still having trouble understanding that (possibly, we may not be agreeing on the definition of computational graph). The number of message passing steps is fixed, once we fix a model. We just need to pass a lot more nodes through the computational graph.

In a naive way:

for each node:
  compute features
  pass to neighbors

You can have several of these pipelines in parallel (i.e. your computational graph has several of these branches to execute for several nodes in parallel), but this parallelism does not depend on the input data. Does it?

Now, there are are issues on how you maintain the graph info. You could compute using sparse like strategies (like the above), or you could use dense strategies with matrix mulitplications. But if you use sparse-like strategies, the size of the computational graph does not change. 

Maybe I am wrong but if I am, I would love to watch a TTT on how pytorch/PhG is expanding this sort of computational graphs.

AugustoPeres:
> > > > > Following @hpenedones comments, is there a way of measuring the I/O performance related with these. I am afraid that all these layers may lead to hidden I/O overheads.
> > > > 
> > > > 
> > > > This is a processing script decoupled from training. Any I/O overhead that happens here will happen only once after processing the data.
> > > > > Unrelated to this: do we already have some info on how the current implementation of GNNs deals with the sparsity of these graphs? I think this was the biggest bottleneck for scaling, right?
> > > > 
> > > > 
> > > > The biggest bottleneck for training is the size of the computational graph used for autograd by pytorch. Which is a consequence of the large graphs in the dataset but not their sparsity
> > > 
> > > 
> > > Can you explain better this:
> > > > The biggest bottleneck for training is the size of the computational graph used for autograd by pytorch. Which is a consequence of the large graphs in the dataset but not their sparsity
> > > 
> > > 
> > > Why would the size of the data graph have impact on the size of the computational graph. Shouldn't the computational graph be determined (mostly) by the number of layers, convolutions, etc?
> > 
> > 
> > It is affected by the number of layers as well. But also by the size of the inputs. Each node that is added to a graph corresponds to more message passing steps, hence larger computational graphs
> 
> I am still having trouble understanding that (possibly, we may not be agreeing on the definition of computational graph). The number of message passing steps is fixed, once we fix a model. We just need to pass a lot more nodes through the computational graph.
> 
> In a naive way:
> 
> for each node: compute features pass to neighbors
> 
> You can have several of these pipelines in parallel (i.e. your computational graph has several of these branches to execute for several nodes in parallel), but this parallelism does not depend on the input data. Does it?
> 
> Now, there are are issues on how you maintain the graph info. You could compute using sparse like strategies (like the above), or you could use dense strategies with matrix mulitplications. But if you use sparse-like strategies, the size of the computational graph does not change.
> 
> Maybe I am wrong but if I am, I would love to watch a TTT on how pytorch/PhG is expanding this sort of computational graphs.

By computational graph I do not mean the architecture of the model. Rather, the Graph that mus be stored in memory for then doing back propagation, i.e:

image

The more operations we do (either by adding layers) or increasing the size of the inputs the more memory will be allocated to these graphs.

At the moment I believe that this is the main issue because I can easily do inference in my personal laptop when using `@torch.no_grad` but when training (where) grad is required we start to have issues

meshnets/process.py
hpenedones: See my other comment: I think that the ML code should receive an HuggingFace dataset already generated (and it should even be done in a separate repo). Because that's what other ML engineers, e.g. outside Inductiva, can directly access. Basically that's the dataset interface to the world.

Therefore, the dataset should make little or no assumptions about the ML models that will consume it (some people might even use MLPs or Convnets or something else). The dataset should just contain all relevant data.

For example, if you want to take the wind speed and pass it as a node feature to every node: what seems more natural is that this would be done in one of the .map operations on the HuggingFace dataset before you pass it to the meshnet. No? But other ML architectures, e.g. concatenating the wind vector to a bottleneck layer of an encoder, would just use it directly, no need to have that in node features.

sarmento:
> > > > > > Following @hpenedones comments, is there a way of measuring the I/O performance related with these. I am afraid that all these layers may lead to hidden I/O overheads.
> > > > > 
> > > > > 
> > > > > This is a processing script decoupled from training. Any I/O overhead that happens here will happen only once after processing the data.
> > > > > > Unrelated to this: do we already have some info on how the current implementation of GNNs deals with the sparsity of these graphs? I think this was the biggest bottleneck for scaling, right?
> > > > > 
> > > > > 
> > > > > The biggest bottleneck for training is the size of the computational graph used for autograd by pytorch. Which is a consequence of the large graphs in the dataset but not their sparsity
> > > > 
> > > > 
> > > > Can you explain better this:
> > > > > The biggest bottleneck for training is the size of the computational graph used for autograd by pytorch. Which is a consequence of the large graphs in the dataset but not their sparsity
> > > > 
> > > > 
> > > > Why would the size of the data graph have impact on the size of the computational graph. Shouldn't the computational graph be determined (mostly) by the number of layers, convolutions, etc?
> > > 
> > > 
> > > It is affected by the number of layers as well. But also by the size of the inputs. Each node that is added to a graph corresponds to more message passing steps, hence larger computational graphs
> > 
> > 
> > I am still having trouble understanding that (possibly, we may not be agreeing on the definition of computational graph). The number of message passing steps is fixed, once we fix a model. We just need to pass a lot more nodes through the computational graph.
> > In a naive way:
> > for each node: compute features pass to neighbors
> > You can have several of these pipelines in parallel (i.e. your computational graph has several of these branches to execute for several nodes in parallel), but this parallelism does not depend on the input data. Does it?
> > Now, there are are issues on how you maintain the graph info. You could compute using sparse like strategies (like the above), or you could use dense strategies with matrix mulitplications. But if you use sparse-like strategies, the size of the computational graph does not change.
> > Maybe I am wrong but if I am, I would love to watch a TTT on how pytorch/PhG is expanding this sort of computational graphs.
> 
> By computational graph I do not mean the architecture of the model. Rather, the Graph that mus be stored in memory for then doing back propagation, i.e:
> 
> image
> 
> The more operations we do (either by adding layers) or increasing the size of the inputs the more memory will be allocated to these graphs.
> 
> At the moment I believe that this is the main issue because I can easily do inference in my personal laptop when using `@torch.no_grad` but when training (where) grad is required we start to have issues

Then we agree on the definition :)
But as you can see, the computational graph does not depend on the *amount* of data given as input: it depends of the format of the input. So, depending on how we represent graphs, this could be a list of nodes, a list of edges and lists attributes, which can be handled in a certain way, and maybe passed x at a time, or it could be a large matrix for the graph, and a few others for attributes (or it could be that the there is such type of conversions inrernally).

LEt's chat about this... :) Live.

sarmento:
> > In fact, one thing we could do to avoid confusion is to have a separate repo, where we have code responsible for generating datasets using our API and packaging them under the HuggingFace interface. And the meshnets repo is focused on training mesh networks but it tries to be a little more agnostic/decoupled from data generation. Wdyt?
> 
> This would be the best option, yes! But since this is a large refactoring maybe we should leave it for a future iteration

Yes. But don't drop this. This seems to be important also for our positioning wrt to the community.

hpenedones:
> > In fact, one thing we could do to avoid confusion is to have a separate repo, where we have code responsible for generating datasets using our API and packaging them under the HuggingFace interface. And the meshnets repo is focused on training mesh networks but it tries to be a little more agnostic/decoupled from data generation. Wdyt?
> 
> This would be the best option, yes! But since this is a large refactoring maybe we should leave it for a future iteration

Yep, sure. Moving to a separate repo later is fine.

meshnets/process.py
hpenedones: Augusto, it is not because we find examples of HuggingFace datasets that hardcode assumptions about the models that consume them, that this is a great decision. We could also, literally, find millions of GitHub repos that contain poor quality code, or that do not follow our programming guidelines, or that are poorly documented, or that don't have unit tests, etc. 

If the data is a graph, the dataset will contain graph data. But graph data can be consumed by other architectures that are not GraphNets (MLPs could use tricks like padding the data to the maximum size, and convnets could deal with varying size graphs, and then have pooling layers to reduce to a smaller, fixed size), and it can be consumed by many different variants of GraphNets.

Therefore, having graph data in the dataset, does not mean we should hardcoded decisions about downstream use by particular models, particularly what is the trick to pass "global" information to the network (e.g. passing as node features everywhere, or just concatenating to a embedding vector in the bottleneck layer of the network). Do you see the difference? What are you going to do when you want to compare different ways of encoding that info in the network, regenerate a dataset in a different format?

AugustoPeres: I will close this pull request, move the generation and post-processing code to a new repo and, while I iterate on that we can schedule a meeting to better discuss exactly what should be returned by the HuggingFace datasets

PR #62, opened 2023-11-06
Adds `to_undirected` function

AugustoPeres:
## Contents

This pull request removes the old processing code from the repository.
Secondly it starts to add the functions to process data from our hugging face datasets. These functions will be used later to map over the dataset and create a graph dataset from the raw data. In particular this first function can be used as follows:

```python
import datasets
from meshenets.data_processing.from_hugging_face import to_undirected

dataset = datasets.load_dataset(''path to data')
dataset = dataset.map(to_undirected)
```

## Testing

This pull request already contains a small testing file. Later, as more functions are built more tests will be implemented.

## `__init__.py` files

The `__init__.py` files were created to allow only the import `from meshnets.data_processing.from_hugging_face import some_function`. Imports with `from meshnets import some_function` or `from meshnets.data_processing import some_function` are not allowed. This will make the repo more future proof. For example if later we decide to use `ray_datasets` we can have the import `from meshnets.data_processing.from_ray import some_function` without any ambiguity.


meshnets/data_processing/from_hugging_face.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def to_undirected(example):
+    """Makes the graph undirected.
+
+    This is equivalent to adding the inverse edges to the graph. That
+    is, if (i, j) is in the `edges` we also add the edge (j, i).
+
+    """
+    edges = np.array(example['edges'])
```
hpenedones, line 29: there is another thing that is implicit here, which is a conversion from a list to a numpy array....

maybe you can decouple that? So your first map function is to_numpy and you convert all the list to numpy arrays.

meshnets/data_processing/test_from_hugging_face.py
```diff
@@ -0,0 +1,23 @@
+"""Test Hugging face processing functions."""
+import numpy as np
+import pytest
+
+from .from_hugging_face import to_undirected
```
hpenedones, line 5: the Python guideline says not to import functions or class names directly... import modules ...

meshnets/data_processing/from_hugging_face.py
hpenedones, line 1: this module name seems a little too specific... and the code does not have any dependency on HuggingFace libraries... it just requires each example to have those fields with lists... Consider calling it data_mappers.py instead...

meshnets/data_processing/test_from_hugging_face.py
hpenedones: this will become: import .data_mappers

hpenedones (commented):
Thanks Augusto, left some small suggestions.

meshnets/data_processing/from_hugging_face.py
AugustoPeres: I understand your comment. Unfortunately that is not how hugging face works. The columns of the Hugging face datasets must be of a very specific set of features, e.g, Array2D, Array3D, text.... Numpy is not one of them, therefore numpy arrays are always represented as lists. As such the `to_numpy` function would be an irrelevant steps.

Take a look at the next few lines of code:

image

hpenedones: Hmm... Interesting....

hpenedones (approved):

meshnets/data_processing/data_mappers.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def to_undirected(example):
+    """Makes the graph undirected.
+
+    This is equivalent to adding the inverse edges to the graph. That
+    is, if (i, j) is in the `edges` we also add the edge (j, i).
+
+    """
+    edges = np.array(example['edges'])
```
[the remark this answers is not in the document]
hpenedones: Yes! Thanks

hpenedones (approved):

AugustoPeres:
By accident merge #64 into this branch instead of main. That is why we now have more files changed when comparing to main. All of these changes have been review and approved. Will merge this to main and everything will be fixed

PR #63, opened 2023-11-07
Adding edge features function

AugustoPeres:
Add the function to compute the the edge features of a given graph.
Adds unit test for the expected result

### DIFF
--- a/meshnets/data_processing/data_mappers.py
+++ b/meshnets/data_processing/data_mappers.py
@@ -32,3 +32,25 @@ def to_undirected(example):
     edges = np.concatenate([edges, edges[:, ::-1]], axis=0)
     example['edges'] = edges
     return example
+
+
+def make_edge_features(example):
+    """Makes the edge features from the node coordinates.
+
+    The edge features:
+
+    - The displacement vector between the source and target nodes.
+    - The distance between the source and target nodes.
+
+    """
+    nodes = np.array(example['nodes'])
+    edges = np.array(example['edges'])
+
+    if edges.size == 0 or nodes.size == 0:
+        example['edge_features'] = []
+        return example
+    displacement_vectors = nodes[edges[:, 1]] - nodes[edges[:, 0]]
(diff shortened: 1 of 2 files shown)


AugustoPeres:
Pylint is failing on `inductiva` API code. I am assuming that there was a new update that broke the code. Nevertheless that code will be removed from the repo entorely on a future pull request:

image

All other actions were successful

AugustoPeres:
> > Pylint is failing on `inductiva` API code. I am assuming that there was a new update that broke the code. Nevertheless that code will be removed from the repo entorely on a future pull request:
> > image
> > All other actions were successful
> 
> That is because, `run_async` is now removed from the API. I guess if you just installed from PyPI you already got a new version, which removes that. All simulations by default run async.

Got it. I will address this in another pull request by removing that code from the repo, thanks

meshnets/data_processing/data_mappers.py
```diff
@@ -32,3 +32,25 @@ def to_undirected(example):
...
+
+    - The displacement vector between the source and target nodes.
+    - The distance between the source and target nodes.
+
+    """
+    nodes = np.array(example['nodes'])
+    edges = np.array(example['edges'])
+
+    if edges.size == 0 or nodes.size == 0:
+        example['edge_features'] = []
```
IvanPombo, line 50: Maybe I am missing something... But why I call the input argument example?

IvanPombo (approved):
Are these functions just for testing purposes?

LGTM, otherwise

meshnets/data_processing/data_mappers.py
AugustoPeres: You are right. I should have given you more context. These functions are to be used to map over our Hugging face datasets. For example:

```python
dataset = datasets.load('wind-tunnel-dataset')
dataset = dataset.map(make_edge_features)
```

This will convert the raw data into whichever format we may want to use to solve the problem.

I called the function argument `example` to follow HuggingFace's [example](https://huggingface.co/docs/datasets/process#map) on mapping over data

PR #65, opened 2023-11-07
Added function to make the node features

AugustoPeres:
Adds a function for node features. At the moment this is just the wind vector.

### DIFF
--- a/meshnets/data_processing/data_mappers.py
+++ b/meshnets/data_processing/data_mappers.py
@@ -66,7 +66,6 @@ def make_node_features(example):
 
     """
     nodes = np.array(example['nodes'])
-    wind_pressure = np.array(example['wind_vector'])
     example['node_features'] = [
         example['wind_vector'] for _ in range(nodes.shape[0])
     ]
--- a/meshnets/data_processing/data_mappers.py
+++ b/meshnets/data_processing/data_mappers.py
@@ -54,3 +54,20 @@ def make_edge_features(example):
     example['edge_features'] = np.concatenate(
         [displacement_vectors, distances[:, None]], axis=1)
     return example
+
+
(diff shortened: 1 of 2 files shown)


meshnets/data_processing/data_mappers.py
```diff
@@ -54,3 +54,19 @@ def make_edge_features(example):
...
+def make_node_features(example):
+    """Makes the node features from the node coordinates.
+
+    The node features:
+
+    - The wind vector. This is a common value for all the nodes in the
+      graph.
+
+    """
+    nodes = np.array(example['nodes'])
```
hpenedones, line 68: why do you need to convert this to numpy? Seems like the only thing you need is the number of nodes to iterate on...  Can you do num_nodes = len(example['nodes']) or similar?

meshnets/data_processing/data_mappers.py
```diff
@@ -54,3 +54,19 @@ def make_edge_features(example):
...
+def make_node_features(example):
+    """Makes the node features from the node coordinates.
+
+    The node features:
+
+    - The wind vector. This is a common value for all the nodes in the
+      graph.
+
+    """
+    nodes = example['nodes']
```
IvanPombo, line 68: Remove nodes here.

IvanPombo (approved):
Otherwise it lgtm!

meshnets/data_processing/data_mappers.py
```diff
@@ -54,3 +54,19 @@ def make_edge_features(example):
...
+    example['node_features'] = [
+        example['wind_vector'] for _ in range(len(nodes))
```
IvanPombo, line 70: If the purpose is to be general, change `wind_vector` with `feature`

PR #67, opened 2023-11-09
Update and remove deprecated ray code

AugustoPeres:
This pull request removes the usage of the `ray_lightning` library that was deprecated earlier this year.
I made only changes to the `train` function in order to remove these dependencies this **was not** meant to be a refactoring of the `train` function itself. That might come later.

## Testing

To test these changes I quickly adapted the scripts `train.py` and `tune.py`. Everything is working fine but I have not commited those files because they have un-merged changes in other pull open requests and I do not want to have to solve conflicts later. As soon as these pull requests are merged I will commit the necessary changes to the scripts.

## More information

https://docs.ray.io/en/latest/train/getting-started-pytorch-lightning.html
https://docs.ray.io/en/latest/train/examples/lightning/lightning_cola_advanced.html#lightning-advanced-example

### DIFF
--- a/meshnets/train.py
+++ b/meshnets/train.py
@@ -9,6 +9,8 @@ from torch.utils.data import random_split
 
 from meshnets.utils import model_training
 from meshnets.utils.datasets import FromDiskGeometricDataset
+from ray.train.torch import TorchTrainer
+from ray.train import ScalingConfig
 
 TUNING_RUN = False
 FLAGS = flags.FLAGS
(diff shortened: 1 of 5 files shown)


AugustoPeres:
As it turns out I do have to commit the changes to the scripts otherwise pylint will fail

meshnets/train.py
```diff
@@ -91,25 +93,29 @@ def main(_):
...
+        'validation_datasets_names': val_datasets_names,
+        'validation_datasets': val_datasets
     }
 
-    model_training.train_model(config,
-                               experiment_config,
-                               train_dataset,
-                               validation_datasets_names=val_datasets_names,
-                               validation_datasets=val_datasets)
+    resources_per_worker = {
```
IvanPombo, line 107: Out of curiosity, this configures our cluster with ray?

meshnets/train.py
```diff
@@ -91,25 +93,29 @@ def main(_):
...
+        'CPU': FLAGS.num_cpus_per_worker,
+        'GPU': 1 if FLAGS.use_gpu else 0
+    }
+    scaling_config = ScalingConfig(num_workers=FLAGS.num_workers_ray,
+                                   use_gpu=FLAGS.use_gpu,
```
IvanPombo, line 112: What is the difference of this part and the one passing the GPU in the resources_per_worker?

meshnets/tune.py
```diff
@@ -92,53 +96,58 @@ def main(_):
...
+        'train_dataset': train_dataset,
+        'validation_datasets_names': val_datasets_names,
+        'validation_datasets': val_datasets
     }
 
+    mlflow.create_experiment(FLAGS.experiment_name)
+
     # Alocate resources per trial.
-    resources_per_trial = {
+    resources_per_worker = {
```
IvanPombo, line 126: What is the difference? Can you update the comment if there is.

meshnets/utils/model_training.py
```diff
@@ -102,14 +87,12 @@ def train_model(config, experiment_config, train_dataset,
...
 
-    train_dataset_name = validation_datasets_names[0]
-    num_params = sum(p.numel() for p in lightning_wrapper.model.parameters())
-
     # The logger creates a new MLFlow run automatically
     # Checkpoints are logged as artifacts at the end of training
     mlf_logger = MLFlowLoggerFinalizeCheckpointer(
         experiment_name=experiment_name)
 
+    num_params = sum(p.numel() for p in lightning_wrapper.model.parameters())
```
IvanPombo, line 95: What does numel do?

IvanPombo (approved):
Just left a couple of minor comments to understand a few things. But it looks good otherwise!

meshnets/train.py
AugustoPeres: This chooses the resources for the function. The cluster is configured outside the script.

We can submit this to the cluster and then it will look for workers capable of meeting the demands.

meshnets/tune.py
AugustoPeres: There really is not much difference. But I will change the names accordingly.

meshnets/utils/model_training.py
AugustoPeres: Returns the total number of elements

meshnets/train.py
AugustoPeres: This chooses the resources for the function. The cluster is configured outside the script.

We can submit this to the cluster and then it will look for workers capable of meeting the demands.

PR #68, opened 2023-11-10
Adds a collat_fn

AugustoPeres:
This pull request adds a collat_fn. This will be important for integrating the hugging face datasets with the torch dataloaders:

```python
>>> import meshnets
>>> import datasets
>>> import torch
>>> import meshnets
>>> dataset = datasets.load_dataset('....', version='....', split='train')
>>> dataset = dataset.map(lambda x: meshnets.data_processing.data_mappers.to_undirected(x)).\
                      map(meshnets.data_processing.data_mappers.make_edge_features).\
                      map(lambda x: meshnets.data_processing.data_mappers.make_node_features(x))
>>> loader = torch.utils.data.DataLoader(dataset, collate_fn=meshnets.data_processing.torch_utils.collate_fn, batch_size=2)
>>> for e in loader: print(e.x.shape)

torch.Size([9466, 3])
torch.Size([9886, 3])
torch.Size([4840, 3])

```

### DIFF
--- a/meshnets/data_processing/torch_utils.py
+++ b/meshnets/data_processing/torch_utils.py
@@ -3,11 +3,11 @@ import torch
 import torch_geometric
 
 
-def collate_fn(examples,
-               x_key='node_features',
-               edge_index_key='edges',
-               edge_attr_key='edge_features',
-               y_key='wind_pressures'):
+def dict_to_geometric_data(examples,
+                           x_key='node_features',
+                           edge_index_key='edges',
+                           edge_attr_key='edge_features',
+                           y_key='wind_pressures'):
     """Collate function for pytorch dataloader
 
     Args:
@@ -20,6 +20,9 @@ def collate_fn(examples,
         features. Defaults to 'edge_features'.
         y_key (str, optional): Key for the target values. Defaults to
         'wind_pressures'.
+
+    Returns:
+        torch_geometric.data.Batch: Batch object containing the
 
     This assumes examples have the structure:
     {
@@ -37,6 +40,18 @@ def collate_fn(examples,
 
     The function wraps the examples in a torch_geometric.data.Data
     object and returns a torch_geometric.data.Batch object.
+
+    python code usage:
+    >>> from torch_geometric.data import DataLoader
+    >>> from torch_geometric.data import Batch
+    >>> from torch_utils import dict_to_geometric_data
+    >>> dataset[0].keys()
+    dict_keys(['node_features', 'edges', 'edge_features', 'wind_pressures'])
+    >>> loader = DataLoader(dataset, batch_size=32,
+    ...                     collate_fn=dict_to_geometric_data_collate)
+    >>> for batch in loader:
+    ...     print(batch)
+    
 
     """
     graphs = [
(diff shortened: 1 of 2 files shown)


meshnets/data_processing/torch_utils.py
```diff
@@ -0,0 +1,17 @@
...
+import torch_geometric
+
+
+def collate_fn(examples,
+               x_key='node_features',
+               edge_index_key='edges',
+               edge_attr_key='edge_features',
+               y_key='wind_pressures'):
+    graphs = [
+        torch_geometric.data.Data(x=torch.tensor(e[x_key]),
```
IvanPombo, line 12: e can mean many things, my first though was edge... Make it for example in examples.

meshnets/data_processing/torch_utils.py
```diff
@@ -0,0 +1,17 @@
+"""Utility functions for usage with pytorch"""
+import torch
+import torch_geometric
+
+
+def collate_fn(examples,
+               x_key='node_features',
+               edge_index_key='edges',
+               edge_attr_key='edge_features',
+               y_key='wind_pressures'):
```
IvanPombo, line 10: Add here a docstring, I don't really know what is a `collate_fn`

IvanPombo (changes requested):
I don't really understand the purpose of this function. Inside it seems that we are just converting of examples into their respective graphs. Why not call it like that? Maybe the reason is because by default this is mentioned as `collate_fn`, however, I am not really familiar with that.. Can you add a comment in the function explaining.

AugustoPeres:
> I don't really understand the purpose of this function. Inside it seems that we are just converting of examples into their respective graphs. Why not call it like that? Maybe the reason is because by default this is mentioned as `collate_fn`, however, I am not really familiar with that.. Can you add a comment in the function explaining.

I called it collate_fn because that is the standard name for the function in pytorch. But yes, this just wraps the examples from the dataset that are of the form: {"nodes": ...} into the `torch_gemoetric` `Data` named tuple used to represent graphs.

Other examples of collat_fn function may include, for example, padding the sequences in a given batch.

Will add a doc string

meshnets/data_processing/torch_utils.py
```diff
@@ -0,0 +1,64 @@
...
+        'node_features'.
+        edge_index_key (str, optional): Key for the edge
+        indices. Defaults to 'edges'.
+        edge_attr_key (str, optional): Key for the edge
+        features. Defaults to 'edge_features'.
+        y_key (str, optional): Key for the target values. Defaults to
+        'wind_pressures'.
+
+    Returns:
+        torch_geometric.data.Batch: Batch object containing the
```
IvanPombo, line 25: the what?

meshnets/data_processing/torch_utils.py
```diff
@@ -0,0 +1,64 @@
...
+
+    This assumes examples have the structure:
+    {
```
IvanPombo, line 28: Should we try to check for this structure then? Or it will through an error immediately?

IvanPombo (approved):
Thanks so much for the update @AugustoPeres! Much better and understandable :D

meshnets/data_processing/torch_utils.py
AugustoPeres: ... graph data. Will this

meshnets/data_processing/torch_utils.py
AugustoPeres: It will fail pretty quickly

PR #70, opened 2023-11-17
Adds a class to compute statistics iterating over the dataset

AugustoPeres:
## Contents

This pull request add back the option to compute the mean and std over the dataset.
For that I implemented a class that updates itself when it receives a new dataset example. It also allows the option to do this operations over a given axis, just like numpy.

## Testing

A pytest file was created comparing the results to numpy.
For an example on how to use the class look also at this test file


meshnets/data_processing/data_statistics.py
```diff
@@ -0,0 +1,79 @@
+"""File with functions to compute data statistics"""
+import numpy as np
+
+
+class OnlineStatistics:
+    """Computes mean ad standart deviation from streaming data
```
hpenedones, line 6: Standard

meshnets/data_processing/data_statistics.py
```diff
@@ -0,0 +1,79 @@
...
+    def update(self, example):
+        data_point = np.expand_dims(example, 0)
+        if self.cumulative_sum is None:
+            self.cumulative_sum = np.sum(data_point,
+                                         axis=self.axis,
+                                         keepdims=True)
+            self.cumulative_sum_2 = np.sum(data_point**2,
+                                           axis=self.axis,
+                                           keepdims=True)
+        else:
```
hpenedones, line 55: DRY principle alert

meshnets/data_processing/data_statistics.py
```diff
@@ -0,0 +1,79 @@
+"""File with functions to compute data statistics"""
+import numpy as np
+
+
+class OnlineStatistics:
+    """Computes mean ad standard deviation from streaming data
```
IvanPombo, line 6: typo *and

meshnets/data_processing/data_statistics.py
```diff
@@ -0,0 +1,79 @@
...
+                                         keepdims=True)
+            self.cumulative_sum_2 = np.sum(np.concatenate(
+                (self.cumulative_sum_2, sum_element_2)),
+                                           axis=self.axis,
+                                           keepdims=True)
+        n = np.sum(np.ones_like(data_point), axis=self.axis, keepdims=True)
+        if self.n is None:
+            self.n = n
+        else:
+            self.n = np.sum(np.concatenate((self.n, n)),
```
IvanPombo, line 70: Should we be able to just add the self.n and n "pointwise"?

meshnets/data_processing/data_statistics.py
```diff
@@ -0,0 +1,79 @@
...
+    array([[0.5], [0.5]])
+    """
+
+    def __init__(self, axis=None):
+        self.axis = axis
+
+        self.n = None
+        self.cumulative_sum = None
+
+        self.cumulative_sum_2 = None
```
IvanPombo, line 44: I think square_sum is a better name for immediate comprehension.

meshnets/data_processing/data_statistics.py
```diff
@@ -0,0 +1,79 @@
...
+        data_point = np.expand_dims(example, 0)
+        self.sum_element = np.sum(data_point, axis=self.axis, keepdims=True)
+        self.sum_element_2 = np.sum(data_point**2,
+                                    axis=self.axis,
+                                    keepdims=True)
+        if self.cumulative_sum is None:
+            self.cumulative_sum = self.sum_element
+            self.cumulative_sum_2 = self.sum_element_2
+        else:
+            sum_element = np.sum(data_point, axis=self.axis, keepdims=True)
```
IvanPombo, line 56: This `sum_element` and `sum_element_2` seem to be exactly the same code you present above with the `self`. So this isn't really required to have here again to thereafter compute the cumulative sums.

IvanPombo (commented):
Left a few comments to improve the clarity of the statistics we want to compute!

meshnets/data_processing/data_statistics.py
AugustoPeres: Not sure what you mean by this

meshnets/data_processing/data_statistics.py
AugustoPeres: Completely agree. Must have missed that when addressing Hugo's comment

meshnets/data_processing/data_statistics.py
IvanPombo: What is the function actually trying to do?

AugustoPeres: When computing the mean of array, in numpy, we can usually do something like:

```python
a = [[1, 2], [3, 4]]
np.mean(a)
# 2.5
np.mean(a, axis=0)
# [[2, 3]]
np.mean(a, axis=1)
# [[1.5], [3.5]]
```

However, our datasets do not fit into memory. So we have to perform this computation online. Thus we keep track of the total sum and the number of samples (`n`).
The extra code is just to keep the functionality of doing this along a given axis (just like in numpy).

We do the same for the standard deviation.

IvanPombo: Just thinking if it is not simple to add pointwise the `n` you computed and the previous one `self.n`. Like, with `np.add`?

AugustoPeres: We could perhaps later improve on the `n` computation. At the moment we also keep track of it using the axis. That is, for the previous example:

```python
s = OnlineStatistics(axis=None)
s.update([1, 2])
s.n
# [2]
s.update([3, 4])
s.n
# [4]
###
s = OnlineStatistics(axis=0)
s.update([1, 2])
s.n
# [[1, 1]]
s.update([3, 4])
s.n
# [[2, 2]]
###
s = OnlineStatistics(axis=1)
s.update([1, 2])
s.n
# [[2]]
s.update([3, 4])
s.n
# [[2], [2]]
```

lneves-ai:
Why not use [npstreams](https://npstreams.readthedocs.io/)?

AugustoPeres:
> Why not use [statstream](https://statstream.readthedocs.io/en/stable/)?

Didn't know that existed. Will take a look and see if we can use it

AugustoPeres:
> Why not use [statstream](https://statstream.readthedocs.io/en/stable/)?

Hi @lneves-ai,

I have tested with npstreams and statstream. The issue is, that they do not allows to compute statistics on iterables that may `yield` arrays of different lengths. Consider for example this dataset:

```python
dataset = [
    [1, 2], # this is element 1 of the dataset
    [1, 2, 3], # this is element 2 of the dataset

]
```

We must be able to compute the mean for each example, that is:

```python
dataset_means = [
    [1.5], # for the element 1 of the dataset
    [2], # for element 2 of the dataset
]
# OR
dataset mean = [[1.8]] # for the whole dataset
```

Examples of different lengths are very frequent on our dataset. However, `np.streams` does not handle this. Consider the following testing code:

```python
def compute_mean(dataset, key, axis, max_iterations=100):
    iterator = make_iterator(dataset, key, max_iterations)
    return npstreams.mean(iterator, axis=axis)

def make_iterator(dataset, key, max_iterations):
    for i in range(max_iterations):
        yield np.array(dataset[i][key])

dataset = [{'nodes': [1, 2]}, {'nodes': [1, 2, 3]}]
print(compute_mean(dataset, 'nodes', axis=1, max_iterations=2)) # Or axis=None
```

It will fail with the error:

```
ValueError: operands could not be broadcast together with shapes (3,) (2,) 

```

With the custom class written here we can perform these operations. I will even push a unit test for this particular case

meshnets/data_processing/data_statistics.py
```diff
@@ -0,0 +1,78 @@
...
+        data_point = np.expand_dims(example, 0)
+        sum_element = np.sum(data_point, axis=self.axis, keepdims=True)
+        sum_element_squared = np.sum(data_point**2,
+                                     axis=self.axis,
+                                     keepdims=True)
+        if self.cumulative_sum is None:
+            self.cumulative_sum = sum_element
+            self.cumulative_sum_squared = sum_element_squared
+        else:
+            self.cumulative_sum = np.sum(np.concatenate(
```
hpenedones, line 57: numpy concatenate of two arrays allocates new memory to create a new bigger array where the data from the two arrays together will be concatenated. You really want to avoid doing this if you just need to compute sums...

meshnets/data_processing/data_statistics.py
```diff
@@ -0,0 +1,78 @@
...
+                (self.cumulative_sum, sum_element)),
+                                         axis=self.axis,
+                                         keepdims=True)
+            self.cumulative_sum_squared = np.sum(np.concatenate(
+                (self.cumulative_sum_squared, sum_element_squared)),
+                                                 axis=self.axis,
+                                                 keepdims=True)
+        n = np.sum(np.ones_like(data_point), axis=self.axis, keepdims=True)
```
hpenedones, line 65: Why would you allocate new memory (with ones_like) and make a np.sum, just extract the size or dimensions of something (i.e. the data_point?)

meshnets/data_processing/data_statistics.py
```diff
@@ -0,0 +1,78 @@
...
+        if self.n is None:
+            self.n = n
+        else:
+            self.n = np.sum(np.concatenate((self.n, n)),
```
hpenedones, line 69: again, please don't use np.concatenate if you just need to compute sums... don't assume that what is logically correct in terms of math operations, makes sense computationally, it can be quite wasteful.

hpenedones (changes requested):

meshnets/data_processing/data_statistics.py
AugustoPeres: I've been thinking about this and, at the moment, could not think of a way to optimize this. Because we are not just computing sums. We are computing sums but we need to respect the axis along which the sums must be computed.

For example, for the dataset:

```python
dataset = [
[1, 2],
[3, 4], 
[5, 6]
]
```

If we want to compute the mean along axis=0 then the cumulative sum is:

```python
# first iteration
[[1, 2]]
# second iteration
[[1, 2]] (+ on axis 0) (sum [[3, 4]] on axis 0) = [[1, 2]] (+ on axis 0) [[3, 4]] = [[4, 6]]
# forth iteration
[[9, 12]]
```

But for axis=1:

```python
# first iteration
[[3]]
# secodn iteration
[[3]] (+ on axis 1) (sum [[3, 4]] on axis 1) = [[3]] (+ on axis 1) [[7]] = [[3], [7]] # Involves concatenation either before or after summing
# third iteration
[[3], [7]] (+ on axis 1) (sum [[4, 6]] on axis 1) = [[3], [7]] (+ on axis 1) [[10]] = [[3], [7], [10]]
```

Which, for both cases a a particular case that, when receiving several samples at the same time the total sum is just `sum_so_far + (sum received samples)` but, for our case, with respect to a given axis which, may involve concatenation either before or after the sum due to the streaming nature of the problem.

I have also tested with `np.stack`, `np.vstack` but they all copy memory as well.
This copying is very small because it duplicates only temporarily one example of the dataset when we may already have thousands in memory.

We could perhaps drop entirely this `streaming_mean` across several dimensions at the same time and focus just on computing the mean for particular cases where we need them. And leave this general case for later if the need arises
For example, if we want to normalize the wind pressures, we can use a normal streaming mean and just iterate over each value of each example on the dataset.

meshnets/data_processing/data_statistics.py
AugustoPeres: Fixed, will commit something soon if it is the case that we want to keep this class: see last paragraphs of https://github.com/inductiva/meshnets/pull/70#discussion_r1402072266

meshnets/data_processing/data_statistics.py
hpenedones: Thanks Augusto for the additional details. I didn't know you wanted to support that case of axis=1 which in fact the "accumulator" keeps growing in size (more elements in the array) --- that seems a little unexpected because the final size will be as long as the total stream size (collapsed along one dimension), so there is not much gain from computing online statistics... Do we need that?

meshnets/data_processing/data_statistics.py
hpenedones: Thanks!

meshnets/data_processing/data_statistics.py
AugustoPeres: We do not in fact need the `axis=1`. I think we can infact go simpler. And just compute the mean of the predictions that we want.

We do need to have a way to do this in a streaming fashion to provide support for datasets in streaming mode

AugustoPeres: I will remove the support for axis=1

hpenedones: Great, thanks!

AugustoPeres:
Closing in favor of #72

PR #72, opened 2023-11-22
Streaming means

AugustoPeres:
Same functionality as #70 but addressing the comments:

1. [Removing support for axis and no need to copy the arrays](https://github.com/inductiva/meshnets/pull/70#discussion_r1401841103)
2. [Use npstreams](https://github.com/inductiva/meshnets/pull/70#issuecomment-1819354342)

Here in fact, the most important part is the `make_iterator`. This function was created so that we can support hugging face datasets in both streaming and non streaming modes


hpenedones (commented):
Great, this is much simpler than the previous implementation! Thanks Augusto!

meshnets/data_processing/dataset_statistics.py
```diff
@@ -0,0 +1,49 @@
...
+
+def make_iterator(dataset, key, axis, max_iterations):
+    """Make iterator over dataset"""
+    if isinstance(dataset, collections.abc.Iterable):
+        for i, example in enumerate(dataset):
+            if i >= max_iterations:
+                break
+            array = np.array(example[key])
+            yield from _yield_array(array, axis)
+    elif isinstance(dataset, collections.abc.Sized):
```
hpenedones, line 24: this one does not support "enumerate"?... there is some code redundancy here...

AugustoPeres: You are rigth. `__len__` is enough to support enumerate

hpenedones (approved):

PR #73, opened 2023-11-24
Add back normalization to training

AugustoPeres:
Adds the dataset normalization to training

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -57,6 +57,10 @@ flags.DEFINE_integer('num_cpus_per_worker', 24,
                      'The number of cpus for each worker.')
 flags.DEFINE_bool('use_gpu', True, 'Whether to use gpu or not.')
 
+flags.DEFINE_integer(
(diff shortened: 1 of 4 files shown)


scripts/train.py
```diff
@@ -57,6 +57,10 @@
                      'The number of cpus for each worker.')
 flags.DEFINE_bool('use_gpu', True, 'Whether to use gpu or not.')
 
+flags.DEFINE_integer(
```
hpenedones, line 60: are the examples shuffled? to avoid that by some reason the first 1000 are all very different from the rest (e.g. they all have the same low wind speed)?

AugustoPeres: The dataset is randomly generated and randomly stored. There is no reason for them not to be shuffled. BUt we can always just use `dataset.shuffle`  provided by the hugging face API

IvanPombo (approved):
LGTM! Could you eventually provided analytics of the dataset to check the distribution of the velocities and pressures?

PR #74, opened 2023-12-06
New dataset map options

AugustoPeres:
This pull request adds `num_proc` and `writer_batch_size` to the HuggingFace datasets. Adding `num_proc` ensures the dataset is mapped in parallel. `writer_batch_size` controls the number of examples written to disk at any one point. The default values for `writer_batch_size` where causing memory issues.

It also moves the code to create the dataloaders to a new function to better reuse these instructions.

### DIFF
--- a/meshnets/utils/model_training.py
+++ b/meshnets/utils/model_training.py
@@ -154,14 +154,10 @@ def train_model(config):
                                     num_workers_loader)
         validation_loaders.append(loader)
 
-    print('Getting node feature size.')
     node_feature_size = get_node_feature_size(train_loader)
-    print('Getting edge feature size.')
     edge_feature_size = get_edge_feature_size(train_loader)
-    print('Getting output size.')
     output_size = get_output_size(train_loader)
 
-    print('Computing dataset statistics.')
     train_stats = compute_train_stats(train_dataset,
                                       config['num_examples_dataset_stats'])
 
@@ -179,7 +175,7 @@ def train_model(config):
         edge_attr_std=train_stats['edge_attr_std'],
         y_mean=train_stats['y_mean'],
         y_std=train_stats['y_std'],
-        validation_datasets_names=val_dataset_versions,
+        validation_datasets_names=[dataset_version] + val_dataset_versions,
         learning_rate=learning_rate)
 
     # The logger creates a new MLFlow run automatically
@@ -187,7 +183,6 @@ def train_model(config):
     mlf_logger = callbacks.MLFlowLoggerFinalizeCheckpointer(
         experiment_name=experiment_name)
 
-    print('counting parameters')
     num_params = sum(p.numel() for p in model.model.parameters())
     # Log the config parameters and training dataset stats for the run to MLFlow
     mlf_logger.log_hyperparams({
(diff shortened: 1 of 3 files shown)


meshnets/utils/model_training.py
```diff
@@ -62,12 +61,59 @@ def compute_train_stats(dataset, max_examples=1000):
...
+    if split_start == 'beginning':
+        split = f'train[:{split_percentage:.0%}]'
+    elif split_start == 'end':
+        split = f'train[-{split_percentage:.0%}:]'
+    else:
+        raise ValueError(f'Invalid split_start: {split_start}')
+    dataset = datasets.load_dataset('inductiva/wind_tunnel',
+                                    version=version,
+                                    split=split,
+                                    download_mode='force_redownload')
```
IvanPombo, line 82: Why this forcing?

meshnets/utils/model_training.py
```diff
@@ -62,12 +61,59 @@ def compute_train_stats(dataset, max_examples=1000):
...
+    dataset = dataset.map(
```
IvanPombo, line 83: For simplicity purposes these next three lines which process the data to create the desired features and structure, should be in a separate method. Wdyt?

IvanPombo: Because this whole function could be also used for other datasets, with different dataset-processing methods.

meshnets/utils/model_training.py
```diff
@@ -90,40 +136,20 @@ def train_model(config):
...
-            data_processing.data_mappers.make_edge_features)
-        val_dataset = val_dataset.map(lambda x: data_processing.data_mappers.
-                                      make_node_features(x, 'wind_vector'))
-
-        validation_loaders.append(
-            torch.utils.data.DataLoader(val_dataset,
-                                        batch_size=batch_size,
-                                        num_workers=num_workers_loader,
-                                        shuffle=False))
+        _, loader = make_dataloader(val_dataset_version, 1, 'beginning',
```
IvanPombo, line 149: Can you comment a bit on this extra validation loaders? I am a bit confused.

IvanPombo (commented):
Nicely done @AugustoPeres! I just left a couple of questions and a simple modular change. Let me know your thoughts!

meshnets/utils/model_training.py
AugustoPeres: To prevent caching issues. Because I have several different versions of the dataset that require different downloads I 'force_download'. Why? because, otherwise HuggingFace will detect `inductiva/wind_tunnel` in `/home/<user>/.cache/huggingface/datasets` and not run the download again. This will cause it to skip the unpack step and may cause it to work with the wrong files

meshnets/utils/model_training.py
AugustoPeres: Not sure I follow. The `map` is a general method of hugging face datasets. But the processing function being used are already problem specific. From datasets following the structure chosen for our HuggingFace dataset they create edge_features, node_features, ...

meshnets/utils/model_training.py
AugustoPeres: > Can you comment a bit on this extra validation loaders? I am a bit confused.

This was a feature implemented by Victor. We have the option of giving other datasets for validation. Then, a validation step for each loader will be created and the metrics `val_loss_{specific_dataset_version}` will be logging to mlflow.
That is, this is just a way to compute performance metrics during the validation step on validation datasets other than the one that results from splitting the training data.

IvanPombo (approved):
Based on the explanations, LGTM!

PR #75, opened 2023-12-06
Adding back visualization features

AugustoPeres:
This pull request adds back the visualization features lost when migrating to Hugging face Datasets. This new function is completely independent of original mesh files and can work only with the examples in the datasets.

Example outputs for an untrained model:
image
image

### DIFF
--- a/meshnets/utils/data_visualization.py
+++ b/meshnets/utils/data_visualization.py
@@ -16,13 +16,16 @@ def plot_3d_graph_and_predictions(example,
                                   save_path=None,
                                   figsize=(12, 6)):
     node_positions = example["nodes"]
+    x = [pos[0] for pos in node_positions]
+    y = [pos[1] for pos in node_positions]
+    z = [pos[2] for pos in node_positions]
 
     fig = plt.figure(figsize=figsize)
 
     ax1 = fig.add_subplot(121, projection="3d")
-    ax1.scatter([pos[0] for pos in node_positions],
-                [pos[1] for pos in node_positions],
-                [pos[2] for pos in node_positions],
+    ax1.scatter(x,
+                y,
+                z,
                 c=example["wind_pressures"],
                 cmap=plt.cm.jet,
                 marker="o",
@@ -33,9 +36,9 @@ def plot_3d_graph_and_predictions(example,
     ax1.set_title("Wind Pressures")
 
     ax2 = fig.add_subplot(122, projection="3d")
-    scatter2 = ax2.scatter([pos[0] for pos in node_positions],
-                           [pos[1] for pos in node_positions],
-                           [pos[2] for pos in node_positions],
+    scatter2 = ax2.scatter(x,
+                           y,
+                           z,
                            c=predictions,
                            cmap=plt.cm.jet,
                            marker="o",
--- a/scripts/visualize.py
+++ b/scripts/visualize.py
@@ -24,6 +24,9 @@ flags.DEFINE_float("train_split", 0.9,
 flags.DEFINE_string("tracking_uri", None,
                     "The tracking URI for the MLFlow experiments.")
 
+flags.DEFINE_string("checkpoint_folder", "checkpoints",
+                    "The folder to load the checkpoint from in mlflow")
+
 flags.DEFINE_string("run_id", None, "The run id of the experiment to load.")
 
 flags.DEFINE_integer("checkpoint", 0,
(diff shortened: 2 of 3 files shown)


scripts/visualize.py
```diff
@@ -43,59 +42,35 @@ def main(_):
...
+        version=FLAGS.dataset_version,
+        split=f"train[-{1 - FLAGS.train_split:.0%}:]",
+        download_mode="force_redownload")
+    len_dataset = len(dataset)
+    random_example = dataset[np.random.randint(len_dataset)]
+    random_example = data_mappers.to_undirected(random_example)
+    random_example = data_mappers.make_edge_features(random_example)
+    random_example = data_mappers.make_node_features(random_example)
+
+    wrapper = model_loading.load_model_from_mlflow(FLAGS.tracking_uri,
```
IvanPombo, line 56: This wrapper refers to the model itself no?

scripts/visualize.py
```diff
@@ -43,59 +42,35 @@ def main(_):
...
-        FLAGS.tracking_uri, FLAGS.run_id, FLAGS.checkpoint)
-
-    sample_idx = np.random.choice(validation_dataset.indices)
-
-    mesh_path = dataset.get_mesh_path(sample_idx)
-    graph_path = dataset.get_graph_path(sample_idx)
+    dataset = datasets.load_dataset(
+        "inductiva/wind_tunnel",
+        version=FLAGS.dataset_version,
+        split=f"train[-{1 - FLAGS.train_split:.0%}:]",
```
IvanPombo, line 48: Are we fetching from the training dataset for the comparisons provided in the PR?

meshnets/utils/data_visualization.py
```diff
@@ -7,6 +10,54 @@
...
+                                  predictions,
+                                  point_size=100,
+                                  save_path=None,
+                                  figsize=(12, 6)):
+    node_positions = example["nodes"]
+
+    fig = plt.figure(figsize=figsize)
+
+    ax1 = fig.add_subplot(121, projection="3d")
+    ax1.scatter([pos[0] for pos in node_positions],
```
IvanPombo, line 23: Was this done like this before?
I recall that Victor was using pyvista for these types of comparisons. However, given that these now may be graphs, it may be simpler to do them with matplotlib.

IvanPombo: This is supposed to work with arrays in general, couldn't just fetch the x, y, z coordinates directly from node_positions?

scripts/visualize.py
```diff
@@ -43,59 +42,35 @@ def main(_):
...
-        dataset, [num_training, size_dataset - num_training])
-
-    wrapper = meshnets.utils.model_loading.load_model_from_mlflow(
-        FLAGS.tracking_uri, FLAGS.run_id, FLAGS.checkpoint)
-
-    sample_idx = np.random.choice(validation_dataset.indices)
-
-    mesh_path = dataset.get_mesh_path(sample_idx)
-    graph_path = dataset.get_graph_path(sample_idx)
+    dataset = datasets.load_dataset(
```
IvanPombo, line 45: Given that you need to do all this preparation, wouldn't be better to create a separate method for this?

scripts/visualize.py
AugustoPeres: Kinda. It is called wrapper because this actually loads the whole `pytorch_lightning` [class](https://github.com/inductiva/meshnets/blob/63fe3db029136f4af6c19dde5a91827e9bf8b576/meshnets/modules/lightning_wrapper.py#L11) that also contains most training.

The itself [receives as argument](https://github.com/inductiva/meshnets/blob/63fe3db029136f4af6c19dde5a91827e9bf8b576/meshnets/modules/lightning_wrapper.py#L15) a pytorch model

scripts/visualize.py
AugustoPeres: At the moment yes. I will generate similar smaller datasets to be used specifically for this validation/visualizations

meshnets/utils/data_visualization.py
AugustoPeres: > Was this done like this before?
No, before we still had the meshes stored in disk. Therefore we where plotting them using pyvisa. Now, because hugging face returns just the arrays represented as graphs we use just matplotlib.

> This is supposed to work with arrays in general, couldn't just fetch the x, y, z coordinates directly from node_positions?

If `node_positions` was an array we could use the [:,column] to extract the positions. However, directly from hugging face we are fed a `list`. For which the syntax does not work

scripts/visualize.py
AugustoPeres: Yes

IvanPombo (approved):
For a first iteration, it looks good to merge!

Let's start considering the ability to log some of these visuals into MLflow or other place. The goal is to save a few samples.

